PR #2, opened 2017-10-26
Add index interface

prateek:
/cc @jeromefroe @robskillington

### DIFF
--- a/doc/types.go
+++ b/doc/types.go
@@ -50,12 +50,8 @@ type Document struct {
 	// ID is the natural identifier for the document.
 	ID ID
 
-	// IndexableFields contains the list of fields by which to index the document.
-	IndexableFields []Field
-
-	// StoredFields contains the list of fields to store supplemental
-	// to the indexable fields.
-	StoredFields []Field
+	// Fields contains the list of fields by which to index the document.
+	Fields []Field
 }
 
 // HashSize is the number of the hashed ID.
--- a/doc/types.go
+++ b/doc/types.go
@@ -1,0 +1,55 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package doc
+
+// ID represents a document's natural identifier.
+type ID []byte
+
+// FieldValueType represents the different types of field values supported.
+type FieldValueType int
+
+const (
+	// UnknownValueType marks un-specified value types.
+	UnknownValueType FieldValueType = iota
+
+	// StringValueType demarks string value types.
+	StringValueType
+)
+
+// Field represents a document field.
+type Field struct {
+	Name      []byte
+	Value     []byte
+	ValueType FieldValueType
+}
+
+// Document represents a document to be indexed.
+type Document interface {
+	// ID is the natural identifier for the document.
+	ID() ID
+
+	// IndexFields contain the list of fields to index by.
+	IndexFields() []Field
+
+	// Bytes returns a serialized representation of the document. It may
+	// comprise contents in addition to the ID/Index fields.
+	Bytes() []byte
+}
--- a/doc/types.go
+++ b/doc/types.go
@@ -40,7 +40,7 @@ type Term []byte
 
 // Field represents a document field.
 // NB(prateek): The value of a Field is a single Term, this deviates
-// from conventional use of the `Term` in other indexing software.
+// from conventional use of `Term` in other indexing software (e.g. Lucene).
 type Field struct {
 	Name      []byte
 	Value     Term
--- a/doc/types.go
+++ b/doc/types.go
@@ -36,6 +36,8 @@ const (
 type Term []byte
 
 // Field represents a document field.
+// NB(prateek): The value of a Field is a single Term, this deviates
+// from conventional use of the `Term` in other indexing software.
 type Field struct {
 	Name      []byte
 	Value     Term
@@ -47,7 +49,7 @@ type Document struct {
 	// ID is the natural identifier for the document.
 	ID ID
 
-	// IndexableFields contain the list of fields by which to index the document.
+	// IndexableFields contains the list of fields by which to index the document.
 	IndexableFields []Field
 
 	// StoredFields contains the list of fields to store supplemental
--- a/doc/types.go
+++ b/doc/types.go
@@ -35,15 +35,13 @@ const (
 	StringValueType
 )
 
-// Term represents the value of a field.
-type Term []byte
+// Value represents the value of a field.
+type Value []byte
 
 // Field represents a document field.
-// NB(prateek): The value of a Field is a single Term, this deviates
-// from conventional use of `Term` in other indexing software (e.g. Lucene).
 type Field struct {
 	Name      []byte
-	Value     Term
+	Value     Value
 	ValueType FieldValueType
 }
 
--- a/doc/types.go
+++ b/doc/types.go
@@ -20,6 +20,8 @@
 
 package doc
 
+import "crypto/md5"
+
 // ID represents a document's natural identifier.
 type ID []byte
 
@@ -34,10 +36,13 @@ const (
 	StringValueType
 )
 
+// Term represents the name/value of a field.
+type Term []byte
+
 // Field represents a document field.
 type Field struct {
-	Name      []byte
-	Value     []byte
+	Name      Term
+	Value     Term
 	ValueType FieldValueType
 }
 
@@ -53,3 +58,12 @@ type Document interface {
 	// comprise contents in addition to the ID/Index fields.
 	Bytes() []byte
 }
+
+// Hash is the hash of a []byte, safe to store in a map.
+type Hash [md5.Size]byte
+
+// IDHashFn computes the Hash for a given document ID.
+type IDHashFn func(id ID) Hash
+
+// TermHashFn computes the Hash for a given term.
+type TermHashFn func(t Term) Hash
--- a/doc/types.go
+++ b/doc/types.go
@@ -20,50 +20,47 @@
 
 package doc
 
-import "crypto/md5"
-
-// ID represents a document's natural identifier.
+// ID represents a document's natural identifier. It's treated as immutable
+// by the indexing subsystem.
 type ID []byte
 
 // FieldValueType represents the different types of field values supported.
-type FieldValueType int
+type FieldValueType byte
 
 const (
-	// UnknownValueType marks un-specified value types.
-	UnknownValueType FieldValueType = iota
-
 	// StringValueType demarks string value types.
-	StringValueType
+	StringValueType FieldValueType = iota
 )
 
-// Term represents the name/value of a field.
+// Term represents the value of a field.
 type Term []byte
 
 // Field represents a document field.
 type Field struct {
-	Name      Term
+	Name      []byte
 	Value     Term
 	ValueType FieldValueType
 }
 
 // Document represents a document to be indexed.
-type Document interface {
+type Document struct {
 	// ID is the natural identifier for the document.
-	ID() ID
+	ID ID
 
-	// IndexFields contain the list of fields to index by.
-	IndexFields() []Field
+	// IndexableFields contain the list of fields by which to index the document.
+	IndexableFields []Field
 
-	// Bytes returns a serialized representation of the document. It may
-	// comprise contents in addition to the ID/Index fields.
-	Bytes() []byte
+	// StoredFields contains the list of fields to store supplemental
+	// to the indexable fields.
+	StoredFields []Field
 }
 
+// HashSize is the number of the hashed ID (currently set at 16 as we're
+// using md5 Hashing).
+const HashSize = 16
+
 // Hash is the hash of a []byte, safe to store in a map.
-type Hash [md5.Size]byte
+type Hash [HashSize]byte
 
-// IDHashFn computes the Hash for a given document ID.
-type IDHashFn func(id ID) Hash
-
-// TermHashFn computes the Hash for a given term.
-type TermHashFn func(t Term) Hash
+// HashFn computes the Hash for a given document ID.
+type HashFn func(id ID) Hash
--- a/doc/types.go
+++ b/doc/types.go
@@ -28,8 +28,11 @@ type ID []byte
 type FieldValueType byte
 
 const (
+	// UnknownValueType demarks an unknown value type.
+	UnknownValueType FieldValueType = iota
+
 	// StringValueType demarks string value types.
-	StringValueType FieldValueType = iota
+	StringValueType
 )
 
 // Term represents the value of a field.
@@ -57,8 +60,7 @@ type Document struct {
 	StoredFields []Field
 }
 
-// HashSize is the number of the hashed ID (currently set at 16 as we're
-// using md5 Hashing).
+// HashSize is the number of the hashed ID.
 const HashSize = 16
 
 // Hash is the hash of a []byte, safe to store in a map.
--- a/index/inmem/types.go
+++ b/index/inmem/types.go
@@ -1,0 +1,91 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package inmem
+
+import (
+	"regexp"
+
+	"github.com/m3db/m3ninx/doc"
+	mi "github.com/m3db/m3ninx/index"
+	"github.com/m3db/m3x/instrument"
+
+	"github.com/RoaringBitmap/roaring"
+)
+
+// Index represents a memory backed index.
+type Index interface {
+	mi.Readable
+	mi.Writable
+
+	// Open prepares the index to accept reads/writes.
+	Open() error
+
+	// Close closes the index.
+	Close() error
+
+	// Freeze makes the index immutable. Any writes after this return error.
+	Freeze() error
+}
+
+type termsDictionary interface {
+	FieldNames() [][]byte
+
+	FetchTerms(FieldName []byte) []doc.Term
+
+	FetchDocs(FieldName []byte, Term doc.Term, negate bool) *roaring.Bitmap
+
+	FetchDocsFuzzy(FieldName []byte, TermRe *regexp.Regexp, negate bool) *roaring.Bitmap
+}
+
+type postingID int64
+
+// postingList represents an efficient mapping from postingID -> bitmap of doc ids
+type postingList interface {
+	// Insert inserts a reference to document `i` for posting id `p`.
+	Insert(p postingID, i mi.DocID)
+
+	// Remove removes the reference (if any) from document `i` from posting id `p`.
+	Remove(p postingID, i mi.DocID)
+
+	// Fetch retrieves all documents which are associated with posting id `p`.
+	Fetch(p postingID) (*roaring.Bitmap, error)
+}
+
+// Options is a collection of knobs for an in-memory index.
+type Options interface {
+	// SetInstrumentOptions sets the instrument options.
+	SetInstrumentOptions(value instrument.Options) Options
+
+	// InstrumentOptions returns the instrument options.
+	InstrumentOptions() instrument.Options
+
+	// SetInsertQueueSize sets the insert queue size.
+	SetInsertQueueSize(value int64) Options
+
+	// InsertQueueSize returns the insert queue size.
+	InsertQueueSize() int64
+
+	// SetNumInsertWorkers sets the number of insert workers.
+	SetNumInsertWorkers(value int) Options
+
+	// NumInsertWorkers returns the number of insert workers.
+	NumInsertWorkers() int
+}
--- a/index/segment/types.go
+++ b/index/segment/types.go
@@ -77,8 +77,8 @@ type Writable interface {
 	// Update updates the given document.
 	Update(d doc.Document) error
 
-	// Delete deletes the given ID.
-	Delete(i doc.ID) error
+	// Delete deletes the given document.
+	Delete(d doc.Document) error
 }
 
 // Writer represents a segment writer.
--- a/index/segment/types.go
+++ b/index/segment/types.go
@@ -1,0 +1,127 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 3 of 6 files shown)


index/types.go
```diff
@@ -0,0 +1,55 @@
...
+	Value []byte
+}
+
+// ID represents a metric identifier.
+type ID []byte
+
+// Index represents a reverse index.
+type Index interface {
+	// Insert inserts the given ID with provided tags.
+	Insert(ID, []Tag) error
```
jeromefroe: Do we want to provide a way to associate data with an `ID`? For instance, for `m3db` we may eventually want to hold metadata about the series in the index. 

Some thoughts I had here are (names need some work):

1. Associate a byte slice with the ID and provide a method to return that data.
```
Insert(ID, []Tag, []byte) error
...
FetchPayload(FetchOptions) []{ID, Payload}
```

2. Take a more generic, document oriented approach.

```
type Document struct {
    ID ID
    Tags []Tag // these are indexed
    Fields []Tag // these are not indexed but are returned
}

Insert(Document) error

Fetch(FetchOptions) []Document
```

index/types.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// Remove removes the given ID.
+	Remove(ID) error
+
+	// Fetch retrieves the list of ids matching the given criterion.
+	Fetch(FetchOptions) []ID
```
jeromefroe: Do we want to be able to return an error here?

index/types.go
```diff
@@ -0,0 +1,55 @@
...
+}
```
jeromefroe: I wonder if it makes sense to have separate `Reader` and `Writer` interfaces? And then combine them in the `Index` interface:

```
type Index interface {
    Writer
    Reader
}
```

My thinking is that it may be beneficial to distinguish when we need to read from the index, write to the index, or both.

index/types.go
prateek: Yep, will do.

index/types.go
prateek: Sure thing. I'll make `Document` an interface with the primitives we need.

jeromefroe: I wonder if it even needs to be an interface, I think a struct might be sufficient.

prateek: The only reason for the interface is to cache additional state (e.g. we'd want to avoid hashing the bytes multiple times)

index/types.go
```diff
@@ -0,0 +1,97 @@
...
+
+// Reader represents an index reader.
+type Reader interface {
+	// Fetch retrieves the list of documents matching the given criterion.
+	Fetch(opts FetchOptions) []doc.Document
+}
+
+// DocID is document identifier internal to each index.
+type DocID uint64
```
jeromefroe: Do we want to use a `uint64` here? From what I can tell most posting list compression methods assume the document ID is 32 bits. For example, [Lucene uses an `int`](https://lucene.apache.org/core/7_1_0/core/index.html) Admittedly this puts a cap of at most 4 billion on the size of the postings lists but I think that's ok as it just makes us support multiple postings lists earlier.

index/types.go
```diff
@@ -0,0 +1,97 @@
...
+
+// PostingList represents an efficient mapping from doc.Term -> []DocID
+type PostingList interface {
+	// Insert inserts a reference to document `i` for term `t`.
```
jeromefroe: Should this be able to return an error? Also should the postings list know anything about the term? I kind of feel it should just work with Doc ID's and the layer above should handle mapping terms to postings list (I also rehash this below, sorry, I'm all jumbled, I wrote the comment below first but it relates to this one also).

index/types.go
```diff
@@ -0,0 +1,97 @@
...
+	Insert(t doc.Term, i DocID)
+
+	// Remove removes the reference (if any) from document `i` to term `t`.
+	Remove(t doc.Term, i DocID)
```
jeromefroe: Hmm, should we able to remove the a value from a `PostingsList`, I think it may be easier to make them append-only and handle tombstones at the layer above.

index/types.go
```diff
@@ -0,0 +1,97 @@
...
+
+	// Terms retrieves all known terms.
```
jeromefroe: I'm not sure these `Terms` methods (and the fetch ones below) belong in the postings list interface. I think they would be better suited on an `Index` or `Segment` interface. I kind of see the postings lists as being really simple and just supports manipulation through opaque doc ids. The `Index` or `Segment` would then have a terms dictionary which is responsible for these lookups. I guess the way I see it is `TermsDictionary` takes term to postings list reference (which may be nil if the term does not exist) and then the postings list can say give you all the doc ids.

index/types.go
```diff
@@ -0,0 +1,97 @@
...
+	Limit           int64
+	SkipID          bool
+	SkipBytes       bool
+	SkipIndexFields bool
+}
+
+// Filter is a field value filter.
+type Filter struct {
+	// FiledName is the name of an index field associated with a document.
+	FieldName []byte
```
jeromefroe: If we take the `Term` definition in the previous file, we could collapse `FieldName` and `FieldValueFilter` into just `Term`.

index/types.go
```diff
@@ -0,0 +1,97 @@
...
+
+package index
+
+import (
+	"regexp"
+
+	"github.com/m3db/m3ninx/doc"
+)
+
+// Writer represents an index writer.
```
jeromefroe: Should we add `Open` and `Close` methods for the `Writer` and `Reader`?

index/types.go
```diff
@@ -0,0 +1,97 @@
...
+type Writer interface {
```
jeromefroe: Do we an `Update` method perhaps? Also, should we expose some way to batch writes together? I imagine we could improve throughput that way.

doc/types.go
```diff
@@ -0,0 +1,55 @@
...
+
+// ID represents a document's natural identifier.
+type ID []byte
+
+// FieldValueType represents the different types of field values supported.
+type FieldValueType int
+
+const (
+	// UnknownValueType marks un-specified value types.
+	UnknownValueType FieldValueType = iota
```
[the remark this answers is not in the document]
prateek: That's fair, I'll change the enum to only support strings for now.

doc/types.go
```diff
@@ -0,0 +1,69 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package doc
+
+import "crypto/md5"
+
+// ID represents a document's natural identifier.
+type ID []byte
```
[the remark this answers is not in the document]
prateek: I put it as a `[]byte` because we can probably save allocs by sharing the underlying bytes (for e.g. when embedding this with m3db).

doc/types.go
```diff
@@ -0,0 +1,69 @@
...
+const (
+	// UnknownValueType marks un-specified value types.
+	UnknownValueType FieldValueType = iota
+
+	// StringValueType demarks string value types.
+	StringValueType
+)
+
+// Term represents the name/value of a field.
+type Term []byte
```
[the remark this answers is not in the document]
prateek: Yep, I modelled it the same way as Lucene. Their [hierarchy](https://lucene.apache.org/core/2_9_4/fileformats.html#Definitions) is Document -> Field -> Term.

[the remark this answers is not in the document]
prateek: The difference being, we only support a single Term per field where as they allow multiple.

doc/types.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Field represents a document field.
+type Field struct {
```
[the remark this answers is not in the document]
prateek: Consider the situation where we have String and Int64 types, the code using a Field looks pretty similar either way:

<details>
<summary> Struct based Field</summary>

```
type Field struct {
 Name Term
 StringValue  Term
 Int64Value int64
 ValueType FieldValueType
}

func processField(f Field) {
  if f.ValueType == StringValueType {
    // do something with f.StringValue 
  } else if f.ValueType == Int64ValueType {
    // do something with f.Int64Value 
  }
}
```
</details>

<details>
<summary> Interface based Field</summary>

```
type Field interface {
 Name() Term
 StringValue()  Term
 Int64Value() int64
 Type() FieldValueType
}

func processField(f Field) {
  if f.Type() == StringValueType {
    // do something with f.StringValue()
  } else if f.Type() == Int64ValueType {
    // do something with f.Int64Value()
  }
}
```
</details>

----- 

Here's the trade-off I see:
- Pro for interface - we could save redundant fields if we had a concrete implementation per value type. E.g. we don't need the space used by int64 value for String field types. 
- Pro for struct - easier evolution - we'd break downstream uses of an interface if we were to add methods, this isn't true for a struct. 

Anything else we should consider before choosing struct v interface for this?

doc/types.go
```diff
@@ -0,0 +1,55 @@
...
+	Name      []byte
+	Value     []byte
+	ValueType FieldValueType
+}
+
+// Document represents a document to be indexed.
+type Document interface {
```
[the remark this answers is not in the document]
prateek: Yeah that's a fair point. All the stuff we talked about for `Field` applies here, with the additional caveat that we'd expect much less changes due to evolution of this type. 

The heap alloc point does make me favour `struct`.

doc/types.go
```diff
@@ -0,0 +1,69 @@
...
+	// IndexFields contain the list of fields to index by.
+	IndexFields() []Field
+
+	// Bytes returns a serialized representation of the document. It may
+	// comprise contents in addition to the ID/Index fields.
+	Bytes() []byte
+}
+
+// Hash is the hash of a []byte, safe to store in a map.
+type Hash [md5.Size]byte
```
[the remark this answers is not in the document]
prateek: Sure, that's fair.

doc/types.go
```diff
@@ -0,0 +1,55 @@
...
+	Value     []byte
+	ValueType FieldValueType
+}
+
+// Document represents a document to be indexed.
+type Document interface {
+	// ID is the natural identifier for the document.
+	ID() ID
+
+	// IndexFields contain the list of fields to index by.
```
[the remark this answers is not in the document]
prateek: `Fields` is a bit too in-specific, we don't want people to put "stored" fields in here too. 

I was trying to avoid putting tense in the name because it implies something about the document's lifecycle. The same document object is in use prior to ingest (which makes `IndexedFields` weird), and once the doc has been ingested (which makes `IndexableFields` weird). Does that make sense?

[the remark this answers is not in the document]
prateek: We should also talk about how we want to handle `StoredFields`

doc/types.go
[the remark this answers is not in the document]
jeromefroe: Yea, kinda sucks that Go doesn't reference immutability. Perhaps its worth calling out that these are immutable then?

doc/types.go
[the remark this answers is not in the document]
jeromefroe: I see, perhaps worth calling out that we only support a single term per field then? Also perhaps we should update `Field` and just make `Value` a `Term`?

doc/types.go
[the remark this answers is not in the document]
jeromefroe: Looks good to me.

index/types.go
prateek: 👍

doc/types.go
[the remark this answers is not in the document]
prateek: talked offline make Name a `[]byte`.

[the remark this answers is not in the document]
prateek: Leaving as struct

doc/types.go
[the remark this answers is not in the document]
prateek: Agreed to change to struct

doc/types.go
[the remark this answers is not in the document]
prateek: Agreed to call it `IndexableFields`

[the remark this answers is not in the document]
prateek: going to stick with Bytes() as the only "stored" field for now. Can model stored fields first class down the road if needed

doc/types.go
```diff
@@ -0,0 +1,69 @@
...
+	IndexFields() []Field
+
+	// Bytes returns a serialized representation of the document. It may
+	// comprise contents in addition to the ID/Index fields.
+	Bytes() []byte
```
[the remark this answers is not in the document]
prateek: As we discussed offline, staying agnostic of this intentionally

index/types.go
prateek: - `Update()` yes
- Batch yes
- will put a todo down for `Commit()` semantics

index/types.go
prateek: yes to both points

index/inmem/types.go
```diff
@@ -0,0 +1,91 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package inmem
```
jeromefroe: Maybe just `mem` or `ram`?

index/types.go
```diff
@@ -0,0 +1,90 @@
...
+	Readable
+
+	// Open prepares the reader to accept reads.
+	Open() error
+
+	// Close closes the reader.
+	Close() error
+}
+
+// DocID is document identifier internal to each index.
```
jeromefroe: Perhaps worth calling out that this limits the size of the indices to 4 billion?

index/inmem/types.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+	"regexp"
+
+	"github.com/m3db/m3ninx/doc"
+	mi "github.com/m3db/m3ninx/index"
```
jeromefroe: Any specific reason to override the package name?

index/types.go
prateek: will do

index/types.go
```diff
@@ -0,0 +1,90 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package index
+
+import (
+	"github.com/m3db/m3ninx/doc"
+)
+
+// Readable is a writable index.
```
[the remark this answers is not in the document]
prateek: will do

index/inmem/types.go
```diff
@@ -0,0 +1,91 @@
...
+// postingList represents an efficient mapping from postingID -> bitmap of doc ids
+type postingList interface {
+	// Insert inserts a reference to document `i` for posting id `p`.
+	Insert(p postingID, i mi.DocID)
+
+	// Remove removes the reference (if any) from document `i` from posting id `p`.
+	Remove(p postingID, i mi.DocID)
+
+	// Fetch retrieves all documents which are associated with posting id `p`.
+	Fetch(p postingID) (*roaring.Bitmap, error)
```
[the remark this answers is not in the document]
prateek: as above, will discuss more in #5

index/inmem/types.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+type termsDictionary interface {
+	FieldNames() [][]byte
+
+	FetchTerms(FieldName []byte) []doc.Term
+
+	FetchDocs(FieldName []byte, Term doc.Term, negate bool) *roaring.Bitmap
+
+	FetchDocsFuzzy(FieldName []byte, TermRe *regexp.Regexp, negate bool) *roaring.Bitmap
```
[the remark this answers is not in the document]
prateek: Sure, will do

index/inmem/types.go
```diff
@@ -0,0 +1,91 @@
...
+	// Freeze makes the index immutable. Any writes after this return error.
+	Freeze() error
+}
+
+type termsDictionary interface {
+	FieldNames() [][]byte
+
+	FetchTerms(FieldName []byte) []doc.Term
+
+	FetchDocs(FieldName []byte, Term doc.Term, negate bool) *roaring.Bitmap
```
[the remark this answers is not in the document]
prateek: Yep, ended up having to do that for locking anyway. Will drop this file and we can discuss in #5

index/inmem/types.go
prateek: Index just felt more verbose. Will revert to fullname for readability.

index/inmem/types.go
prateek: `mem` it is

doc/types.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// IndexableFields contain the list of fields by which to index the document.
+	IndexableFields []Field
+
+	// StoredFields contains the list of fields to store supplemental
+	// to the indexable fields.
+	StoredFields []Field
+}
+
+// HashSize is the number of the hashed ID (currently set at 16 as we're
```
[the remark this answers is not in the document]
prateek: ATM using it the key for an internal 'don't write' cache

doc/types.go
```diff
@@ -0,0 +1,66 @@
...
+
+const (
+	// StringValueType demarks string value types.
+	StringValueType FieldValueType = iota
+)
+
+// Term represents the value of a field.
+type Term []byte
+
+// Field represents a document field.
```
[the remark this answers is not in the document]
prateek: Sure will do

doc/types.go
```diff
@@ -0,0 +1,66 @@
...
+// ID represents a document's natural identifier. It's treated as immutable
+// by the indexing subsystem.
+type ID []byte
+
+// FieldValueType represents the different types of field values supported.
+type FieldValueType byte
+
+const (
+	// StringValueType demarks string value types.
+	StringValueType FieldValueType = iota
```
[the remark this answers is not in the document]
prateek: That's unidiomatic to me. Either we have string as the default, or an unknown as the default. Shouldn't skip assigning `iota` either way.

doc/types.go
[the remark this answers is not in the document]
prateek: done

doc/types.go
```diff
@@ -0,0 +1,66 @@
...
+	Value     Term
+	ValueType FieldValueType
+}
+
+// Document represents a document to be indexed.
+type Document struct {
+	// ID is the natural identifier for the document.
+	ID ID
+
+	// IndexableFields contain the list of fields by which to index the document.
```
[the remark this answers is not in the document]
prateek: done

doc/types.go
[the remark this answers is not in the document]
prateek: done

index/inmem/types.go
prateek: done

index/inmem/types.go
prateek: done

index/inmem/types.go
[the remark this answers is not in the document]
prateek: done

index/inmem/types.go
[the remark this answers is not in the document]
prateek: dropped, will followup in next pr

index/inmem/types.go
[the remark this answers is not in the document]
prateek: done

index/types.go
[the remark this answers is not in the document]
prateek: done

index/types.go
prateek: done

doc/types.go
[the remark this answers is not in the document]
jeromefroe: In that case, I feel that we should add the unknown back. I'd prefer to not have the uninitialized value be valid.

doc/types.go
```diff
@@ -0,0 +1,68 @@
...
+	// StringValueType demarks string value types.
+	StringValueType FieldValueType = iota
+)
+
+// Term represents the value of a field.
+type Term []byte
+
+// Field represents a document field.
+// NB(prateek): The value of a Field is a single Term, this deviates
+// from conventional use of the `Term` in other indexing software.
```
jeromefroe: nit: /of the Term/of Term

doc/types.go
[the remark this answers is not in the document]
jeromefroe: Perhaps remove the reference to md5 then since its an implementation detail?

doc/types.go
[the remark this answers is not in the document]
prateek: Sure

doc/types.go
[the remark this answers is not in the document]
prateek: Sure

doc/types.go
prateek: done

doc/types.go
```diff
@@ -0,0 +1,70 @@
...
+const (
+	// UnknownValueType demarks an unknown value type.
+	UnknownValueType FieldValueType = iota
+
+	// StringValueType demarks string value types.
+	StringValueType
+)
+
+// Term represents the value of a field.
+type Term []byte
```
jeromefroe: Was thinking about this some more, given that `Term` is often used to expressed a field and its value, maybe just `Value` would be more appropriate?

doc/types.go
```diff
@@ -0,0 +1,70 @@
...
+	ValueType FieldValueType
+}
+
+// Document represents a document to be indexed.
+type Document struct {
+	// ID is the natural identifier for the document.
+	ID ID
+
+	// IndexableFields contains the list of fields by which to index the document.
+	IndexableFields []Field
```
jeromefroe: Was looking through some Lucene documentation some more and they actually distinguish three properties of fields: indexed, analyzed (we don't need to worry about this for now, though it can't hurt to think about) and stored. Indexed determines only whether a field is indexed or not, so whether that field is searchable. Stored determines if the field is stored in the index, so whether or not it can be returned in a query. These configurations are made at the field level though across the entire index, not at the individual document level. So perhaps a document should just contain `Fields` and the determination of whether it will be indexed and/or stored is made in the `IndexWriter`?

index/segment/types.go
```diff
@@ -0,0 +1,127 @@
...
+
+// Iter allows for iteration over the documents indexed in the segment.
+type Iter interface {
+	// Next returns a bool indicating if the iterator has any more documents
+	// to return.
+	Next() bool
+
+	// Current returns the current document and the DocID associated with the
+	// document in the segment.
+	Current() (doc.Document, DocID)
```
jeromefroe: How do you feel about `Value` instead of `Next`? Just seems like it might be more natural, since `Next` and `Current` are very similar but one returns a bool and the other the values.

index/segment/types.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// Err returns any errors the Iter ran into whilst iterating.
+	Err() error
+}
+
+// Readable is a readable index segment.
+type Readable interface {
+	// Fetch retrieves the list of documents matching the given criterion.
+	Fetch(opts FetchOptions) []doc.Document
```
jeromefroe: Instead of passing options, perhaps we should pass a query?

index/segment/types.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// Writable is a writable index segment.
+type Writable interface {
+	// Insert inserts the given documents with provided fields.
+	Insert(d doc.Document) error
+
+	// Update updates the given document.
+	Update(d doc.Document) error
```
jeromefroe: Don't we need the ID here?

index/segment/types.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// Close closes the reader.
+	Close() error
+}
+
+// DocID is document identifier internal to each index segment. It's limited to
+// be at most 2^32.
+type DocID uint32
+
+// FetchOptions is a group of criterion to filter documents by.
```
jeromefroe: Same as above, maybe `Query` would be more appropriate?

index/segment/types.go
```diff
@@ -0,0 +1,127 @@
...
+type FetchOptions struct {
+	IndexFieldFilters []Filter
```
jeromefroe: Hmm, I'm not sure Filters is always appropriate, especially if we want to support conjunction queries. What about using `Terms` here?

index/segment/types.go
```diff
@@ -0,0 +1,127 @@
...
+// ReservedFieldName signifies the field names which are reserved for internal
+// use.
+type ReservedFieldName string
+
+const (
+	// IDFieldName is the fieldname used for the `doc/Document.ID` field.
+	IDFieldName ReservedFieldName = "_id"
+)
+
+// ID is a unique id per segment.
```
jeromefroe: What's this for? To distinguish segments?

doc/types.go
prateek: Sure

index/segment/types.go
prateek: Named like this to follow the convention used in m3db (e.g. [here](https://github.com/m3db/m3db/blob/master/encoding/types.go#L107-L121))

index/segment/types.go
prateek: Sure, I'll rework to have better compostability

index/segment/types.go
prateek: Do you mean doc.ID (internal) or Document.ID(external)? atm, we're not exposing the internal ID to users (outside of the Iter()) method, so a delete is the same as an insert w/ a tombstone marker

index/segment/types.go
prateek: Yep.

doc/types.go
prateek: Re: the distinction of stored & indexed fields at Doc v Index - effectively comes down to if we want to make the Index schema aware or not. If we treat all fields in the document as the same, we'd have to make the Index aware of which ones we'd need to index. Our underlying problem is people can add whatever tag name they'd like for a metric, so by default all fields would have to be indexed, and some  'known' ones per Index would be overriden to be only stored. It feels cleaner to have the distinction at the document level instead of some scheme like that. 

Re analyzed fields: Analysers/Tokenisers add a lot of complexity. We can come back to this if ever need them.

jeromefroe: In regards to supporting a schema or not, I actually think it would be preferable to have defaults for the fields with overrides setup at initialization for some. If a field is indexed for one document but not for others I think that can lead to confusing results.

Not suggesting we add Analysis at this stage :)

index/segment/types.go
jeromefroe: That's right, disregard this, I forgot the ID is contained in the doc already.

doc/types.go
prateek: done

index/segment/types.go
prateek: fixed

index/segment/types.go
prateek: take a look at the new struct and see if it addresses this

doc/types.go
prateek: Either way the burden to specify what to index lies on the user. I'd prefer going the schema aware route too but I'm concerned about indexing redundant fields. 

Considering that metric tag names can being anything, we have to index all fields by default. Say everything's working fine and now we want to tack on a stored field to be returned with a metric  (e.g. an extra R2 policy). People are just going to add that field to the document and move on. Adds a lot of work for them to go make the change to the index to say all fields with this name are stored fields, etc. It's much easier to encourage the correct behaviour by allowing them to specify index v stored at the document level. 

TBH, I'd change my tune if we could change the default to be un-indexed. At that point, it'd be much better for the system to be schema aware at the Index level.

jeromefroe: I think Elasticsearch handles pretty cleanly already. When an index is created you specify a template which can include fields and their attributes and also defaults. Then when you index a document, if it has a field in the schema, the attributes for that schema are applied. Otherwise a new field is created in the schema and the defaults are applied. New fields can added to the schema after the fact as long as they haven't been added to the schema already.

prateek: I don't think this something we need to decide right now. Lets start by supporting only indexed fields, i.e. no stored fields. I'll change `Document` to the following definition:

```
type Document interface {
  ID ID 
  Fields []Field
}
```

We can come back to this discussion when we want to support stored fields.

jeromefroe: Sounds good to me. Just to be clear though, I think whether a field is stored and whether it is indexed are orthogonal concepts. A field can be
- (stored, indexed) in which case it can be searched for and returned in results
- (not stored, indexed) in which case it can be searched for but not returned in results, I see this as being something like the contents of the book, you might store metadata about the book and a path to where it can be found, but there's probably no need to include the contents of the book in an index itself if it can be found elsewhere when needed
- (stored, not indexed) this might be metadata that would never be searched for but is helpful to see in results
- (not stored, not indexed) this is just weird

Anyway, just wanted to clarify my own thoughts. Starting with every field being (stored, indexed) sounds good to me.

PR #3, opened 2017-10-27
[bench] Add types for benchmarking Prometheus remote storage

jeromefroe:
This diff adds the types for benchmarking a Prometheus remote storage endpoint. A subsequent diff will add actual implementations of the interfaces.

cc @prateek

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -1,0 +1,7 @@
+package: github.com/m3db/m3index
+import:
+- package: github.com/prometheus/prometheus
+  # TODO(jeromefroe): When the new Prometheus remote storage endpoint is included
+  # in a stable release we should use a version here instead of a commit SHA.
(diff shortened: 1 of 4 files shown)


glide.yaml
```diff
@@ -0,0 +1,7 @@
+package: github.com/m3db/m3index
+import:
+- package: github.com/prometheus/prometheus
+  # TODO(jeromefroe): When the new Prometheus remote storage endpoint is included
+  # in a stable release we should use a version here instead of a commit SHA.
+  version: ee011d906d683278518fa3d8439d08b0871e93ff
+- package: github.com/tylertreat/bench
```
prateek: maybe pin this too

jeromefroe: Yea, there aren't any releases so I figured I would just leave it unpinned so that on the off chance another package we import as well also uses it we won't potentially clash. That's unlikey though, I'll just pin it.

PR #5, opened 2017-11-02
mem.Segment implementation using 2-level map for terms dict

prateek:
Included in this PR:
- interface + implementation for `PostingsList` backed by roaring.Bitmap 
- interface + implementation for `termsDictionary` using a 2-level map
- implementation for `mem.Segment ` using the primitives above
- variety of tests


index/inmem/index.go
```diff
@@ -0,0 +1,297 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package inmem
+
+import (
+	"crypto/md5"
+	"sync"
+
+	"github.com/RoaringBitmap/roaring"
```
prateek: address import order

index/inmem/index.go
```diff
@@ -0,0 +1,297 @@
...
+)
+
+const initialSize = 1024 * 1024
+
+// TODO(prateek): investigate impact of native heap
+type index struct {
+	opts   Options
+	hashFn doc.HashFn
+
+	docIDGen *atomic.Uint64
```
prateek: int32

index/inmem/postings_list.go
```diff
@@ -0,0 +1,171 @@
...
+type postingsList interface {
+	// InsertNew inserts document `i` at a new posting id, and returns the
+	// positing id used.
+	InsertNew(i mi.DocID) postingID
+
+	// Insert inserts a reference to document `i` for posting id `p`.
+	Insert(p postingID, i mi.DocID)
+
+	// Remove removes the reference (if any) from document `i` from posting id `p`.
+	Remove(p postingID, i mi.DocID)
```
jeromefroe: Do we want to support `Remove`? This would essentially exclude us from using delta encoding which is the standard approach.

index/inmem/postings_list.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/uber-go/atomic"
+)
+
+type postingID int
+
+// postingsList represents an efficient mapping from postingID -> bitmap of doc ids
+type postingsList interface {
+	// InsertNew inserts document `i` at a new posting id, and returns the
+	// positing id used.
+	InsertNew(i mi.DocID) postingID
```
jeromefroe: Any reason to have a special `InsertNew` function? Seems a little out of place.

index/inmem/index.go
```diff
@@ -0,0 +1,297 @@
...
+	panic("not implemented")
+}
+
+func (i *index) Merge(...mi.Readable) error {
+	panic("not implemented")
+}
+
+type termsDict struct {
+	sync.RWMutex
+	terms map[string]postingID
```
jeromefroe: What's a posting ID? Why not just have a reference to the postings list?

index/inmem/postings_list.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Insert inserts a reference to document `i` for posting id `p`.
+	Insert(p postingID, i mi.DocID)
```
jeromefroe: Had a similar comment above, but do we need a postings ID? Why not associate the postings list with the term directly?

index/inmem/postings_list.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (d *docIDs) clone() *roaring.Bitmap {
+	d.RLock()
+	clone := d.bitmap.Clone()
+	d.RUnlock()
+	return clone
+}
+
+func (d *docIDs) andTo(b *roaring.Bitmap) {
```
jeromefroe: perhaps `intersect` would be clearer?

index/inmem/postings_list.go
```diff
@@ -0,0 +1,171 @@
...
+	Insert(p postingID, i mi.DocID)
+
+	// Remove removes the reference (if any) from document `i` from posting id `p`.
+	Remove(p postingID, i mi.DocID)
+
+	// Fetch retrieves all documents which are associated with posting id `p`.
+	Fetch(p postingID) *docIDs
+}
+
+type pList struct {
```
jeromefroe: perhaps `roaringPostingsList` would make it clear that this implementation uses Roaring bitmaps?

index/mem/index.go
```diff
@@ -0,0 +1,302 @@
...
+)
+
+const initialSize = 1024 * 1024
+
+// TODO(prateek): investigate impact of native heap
+type memIndex struct {
+	opts   Options
+	hashFn doc.HashFn
+
+	docIDGen *atomic.Uint32
```
jeromefroe: Should this be per shard?

index/mem/index.go
```diff
@@ -0,0 +1,302 @@
...
+
+	// hash(doc.ID)  -> internal docID
+	idsLock sync.RWMutex
+	idsMap  map[doc.Hash]index.DocID
```
jeromefroe: Is it perhaps worthwhile to make this a two-level map? The first level being just like it is now, hash(doc.ID) to a struct with the internal DocID and the doc.ID, the second level then being doc.ID to internal DocID. This way we can guarantee that even if two id's have the same hash we're safe. Also, we can probably reduce the size of the hash to 64 bits since we have a failsafe.

index/mem/index.go
```diff
@@ -0,0 +1,302 @@
...
+
+	// internal docID -> document
+	documentsLock sync.RWMutex
+	documents     map[index.DocID]document
+
+	// fieldName -> (Term -> postingID)
+	fieldLock sync.RWMutex
+	fieldMap  map[string]*termsDict
```
jeromefroe: Is it perhaps worth wrapping these in interfaces so we can easily swap out the maps for another implementation later on? For example, I'd imagine we may want to explore sync.Map as we increase concurrency.

index/mem/index.go
```diff
@@ -0,0 +1,302 @@
...
+
+	// postingID -> bitmap of docIDs
+	postingsList postingsList
+}
+
+type document struct {
+	doc.Document
+	docID      index.DocID
+	tombstoned bool
```
jeromefroe: Do we want to index a document when it is deleted? Perhaps it would be preferable to instead keep a bitmap or postings list of deleted documents in the index, and then when we get the postings list of documents that match the query, we take the set difference with that. Seems to be the [approach taken by Lucene](https://www.elastic.co/blog/lucenes-handling-of-deleted-documents).

index/mem/index.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// New returns a new in-memory index.
+func New(opts Options) Index {
+	hashFn := func(i doc.ID) doc.Hash {
+		return md5.Sum([]byte(i))
```
jeromefroe: Perhaps an over optimization, but maybe its worth using a different non-cryptographic hash function? I did [some benchmarks](https://github.com/jeromefroe/golang_benchmarks#non-cryptographic-hash-functions) a while back and md5 is one of the slowest non-cryptographic hash while not offering any benefits in desirable non-cryptographic hash function properties (even distribution, avalanche effect, etc.)

index/mem/index.go
```diff
@@ -0,0 +1,302 @@
...
+
+// TODO(prateek): consider copy semantics for input data, esp if we can store it on the native heap
+func (i *memIndex) Insert(d doc.Document) error {
+	// early terminate if id is already present
+	docHash := i.hashFn(d.ID)
+	i.idsLock.RLock()
+	_, ok := i.idsMap[docHash]
+	i.idsLock.RUnlock()
+	if ok {
+		return nil
```
jeromefroe: I think we should return an error here. Although the ID is the same, the contents of the document may be different from what is already stored in the index and the caller would have no way of knowing that we didn't store the contents of the document if they did change.

index/mem/index.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	// insert into document id map
+	i.idsLock.Lock()
+	// check if it's been inserted since we released lock
+	_, ok = i.idsMap[docHash]
+	if ok {
+		i.idsLock.Unlock()
+		return nil
```
jeromefroe: Same as above regarding returning an error.

index/mem/index.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+
+		if !docsSet {
+			candidateDocIds = fetchedIds.clone()
+			docsSet = true
+			continue
+		}
+
+		// update candidate set
+		fetchedIds.andTo(candidateDocIds)
```
jeromefroe: Instead of intersecting one by one, perhaps its preferable to wait until we've extracted all IDs? Then we can optimize the intersection. For example, if any our empty we can just return without actually doing any intersections. Barring that, we can then perform the intersections in size order, smallest first.

index/mem/index.go
```diff
@@ -0,0 +1,302 @@
...
+	panic("not implemented")
+}
+
+func (i *memIndex) Merge(...index.Readable) error {
+	panic("not implemented")
+}
+
+type termsDict struct {
+	sync.RWMutex
+	terms map[string]postingID
```
jeromefroe: What's a posting ID? Why not just have a reference to the postings list?

index/mem/postings_list.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/uber-go/atomic"
+)
+
+type postingID int
+
+// postingsList represents an efficient mapping from postingID -> bitmap of doc ids
+type postingsList interface {
+	// InsertNew inserts document `i` at a new posting id, and returns the
+	// positing id used.
+	InsertNew(i index.DocID) postingID
```
jeromefroe: Any reason to have a special InsertNew function? Seems a little out of place.

index/mem/postings_list.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Insert inserts a reference to document `i` for posting id `p`.
+	Insert(p postingID, i index.DocID)
```
jeromefroe: Had a similar comment above, but do we need a postings ID? Why not associate the postings list with the term directly?

index/mem/postings_list.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Remove removes the reference (if any) from document `i` from posting id `p`.
+	Remove(p postingID, i index.DocID)
```
jeromefroe: Do we want to support Remove? This would essentially exclude us from using delta encoding which is the standard approach.

index/mem/postings_list.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Fetch retrieves all documents which are associated with posting id `p`.
+	Fetch(p postingID) *docIDs
+}
+
+type pList struct {
```
jeromefroe: perhaps `roaringPostingsList` would make it clear that this implementation uses Roaring bitmaps?

index/mem/postings_list.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (d *docIDs) clone() *roaring.Bitmap {
+	d.RLock()
+	clone := d.bitmap.Clone()
+	d.RUnlock()
+	return clone
+}
+
+func (d *docIDs) andTo(b *roaring.Bitmap) {
```
jeromefroe: perhaps `intersect` would be clearer?

doc/metric.go
```diff
@@ -0,0 +1,37 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package doc
+
+// NewMetric returns a new metric document.
+func NewMetric(id []byte, tags map[string]string) Document {
```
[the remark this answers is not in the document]
prateek: > Any reason to not make this a more generic NewDoc function instead?

Sure, I'll rename. 

> Perhaps the value type of the map should be []byte?

I'd like to eventually structure it so it doesn't require any allocs. So we might end up changing the signature to something like 

` func NewDoc(id []byte, tagKeys [][]byte, tagValues [][]byte) Document {... `

I'll put a TODO down for now, we can revisit once we see how inter-op with m3db works. 

> Is it worth calling out that field names must be valid UTF-8?

Hmm yeah that's a fair point. I'll put it in the same note as above. 

> Also, it just occurred to me that we'll probably want to index the ID's so perhaps we to have some reserved field names? Or maybe we just accept a map of fields and if _id is set then we use that as the ID, otherwise we generate an ID ourselves? FWIW this seems to be the approach taken by Elasticsearch

I was going the route of being more strict about this. `ID` has to be set, and it's the unique identifier for a document. The internal doc id we assign to a document won't matter to users of the API.

index/mem/index.go
prateek: Nah, it'll be per `index.Writable`/`index.Readable`. We'd handle collisions when merging indices

doc/metric.go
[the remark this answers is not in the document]
jeromefroe: All sound good to me, but perhaps we should have `NewDoc` and `NewDocWithID` with the former being a convenience wrapper for the latter where we generate the ID ourselves?

index/mem/index.go
jeromefroe: How will we combine results from multiple indices then? It will have to be by Doc.ID which will be much more expensive.

doc/metric.go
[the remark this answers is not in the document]
prateek: Thought about the `_id` suggestion some more, I like it. 

Re: generating the ID, I don't think the index should do that. If people don't have an ID that fits into their natural scheme, I'd prefer they generated the `UUID` or w/e we'd end up doing internally. That way they're able to define uniqueness for their use-case.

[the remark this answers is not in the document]
jeromefroe: Yea, I just feel like for most use cases there won't in fact be a unique ID. It's easy to add such a thing in the future though if the need arises.

index/mem/index.go
prateek: Sure thing

index/mem/index.go
prateek: Agreed. Will do.

index/mem/index.go
prateek: Will do.

index/mem/index.go
prateek: If we don't treat this per index, we'd be saying an index shard cannot index more 4B unique metrics over it's lifetime (both memory + filesystem). That's a big architectural constraint we're enforcing when we don't need to. Plus it's more state to persist. 

I do see your point about the cost of the fetch. We can do some work to make the multi-index fetch more efficient, eg hash-merge or merge-sort in the worst case. 

Not ideal but the better trade-off no?

index/mem/index.go
prateek: tl;dr ~ yeah we need to, will put down a TODO. 

I'm using the map for two things:
1) as a cheap "don't write cache",
2) for cheaper updates because we were able to do a direct lookup by doc hash. 

For (1) I'd actually like this to be a LRUCache so we can treat it as a reverse bloom-filter. I'll put down a TODO for this too. 

For (2) I'm getting more inclined to treat all updates as a straight insert. Even the in-memory ones. Again handling the dedupe read side. 

p.s. (2) made me realise we're going to need some kind of `Optimize()` call to cleanup dead/redundant doc references. Will add that to the `Writable` interface.

index/mem/index.go
prateek: Sure, i'll put down a TODO to evaluate perf impact of different hash functions

index/mem/index.go
prateek: Aren't those separate things? 

The deleted DocID/bitmap would be unique per segment. We'd need to have a tombstone indicator per document id to be able to merge deletes across segments. 

Keeping a bitmap per segment would def help, will put down a TODO

index/mem/index.go
prateek: There's a bunch of ways to optimise this. For e.g. if we are applying 10 filters which eventually lead to no data, but take a long time to retrieve respective doc lists then early application can helpful in pipelining cpu cost. 

I'll create an interface where we can accumulate `fetchedIds`. Something like `docFetcher`, which'll make the decisions based on w/e heuristics we come up with.

index/mem/postings_list.go
prateek: This posting list is only for the `mem` package, I imagine the `fs` package will have it's own.

index/mem/postings_list.go
prateek: sure

index/mem/postings_list.go
prateek: I didn't want to use a name implying the receiver was going to be changed, how about `project` instead?

doc/metric.go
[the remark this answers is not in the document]
prateek: done

index/mem/index.go
prateek: Moved this map to the struct backing the `Index` interface, so it can be shared amongst all segments. Left a todo explaining this discussion.

index/mem/index.go
prateek: done

index/mem/index.go
prateek: done

index/mem/index.go
prateek: done

index/mem/index.go
prateek: done at the Index level (instead of at the Segment level)

index/mem/index.go
prateek: done

doc/doc.go
```diff
@@ -0,0 +1,39 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package doc
+
+// New returns a new document.
+// TODO: figure out if we can get away with fewer allocs, e.g. we could update
+// the signature to be New(id []byte, tagKeys [][]byte, tagValues [][]byte) Document {...}
```
jeromefroe: To that end, perhaps we can just take a slice of `Fields` and push it onto the caller to guarantee that the names are unique?

index/segment/mem/index.go
```diff
@@ -0,0 +1,225 @@
...
+func New(opts Options) Segment {
+	hashFn := func(i doc.ID) doc.Hash {
+		return md5.Sum([]byte(i))
+	}
+
+	return &memIndex{
+		opts:   opts,
+		hashFn: hashFn,
+
+		docIDGen: atomic.NewUint32(0),
```
jeromefroe: Instead of making this per-segment, perhaps it would be preferable to make it a `uint64` that is per index? By making it per-index we won't have to merge the stored fields when we merge segments since we can store them separately from the segments themselves. Deleting documents should be cheaper as well since we would only need to store a bitmap of the deleted document IDs.

index/segment/mem/postings_list.go
```diff
@@ -0,0 +1,171 @@
...
+type postingsList interface {
+	// InsertNew inserts document `i` at a new posting id, and returns the
+	// positing id used.
+	InsertNew(i segment.DocID) postingID
+
+	// Insert inserts a reference to document `i` for posting id `p`.
+	Insert(p postingID, i segment.DocID)
+
+	// Remove removes the reference (if any) from document `i` from posting id `p`.
+	Remove(p postingID, i segment.DocID)
```
jeromefroe: This would be hard to support if we want to use delta-encoding? Perhaps we should strictly handle deletes through merges? I also can't imagine we'll often receive deletes for a document soon after it is created. Also, if we don't include `Remove` we can reuse the postings list interface and perhaps implementation between mem and fs.

index/segment/mem/postings_list.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (d *docIDs) clone() *roaring.Bitmap {
+	d.RLock()
+	clone := d.bitmap.Clone()
+	d.RUnlock()
+	return clone
+}
+
+func (d *docIDs) andTo(b *roaring.Bitmap) {
```
jeromefroe: `intersect` seems clearer to me, what do you think?

doc/doc.go
prateek: That's definitely the intent. Don't want to take a slice of fields cause that's the whole point of this method. If they had that, they could create a `Document` anyway (it's exported after all).

index/segment/mem/index.go
prateek: The deleted documents benefit is fair, but consider that users don't have the internal document id when specifying a delete. We're going to have to handle delete compactions in addition to this anyway. Same thing for stored fields, regardless of internal doc id, we'd need to compact across the same external doc id. 

I have a few more concerns with this:
- it'd add contention to the write path (e.g. if we ran multiple mem segments to increase write throughput, they'd all compete for this resource)
- it'd increase the memory footprint of the bitmaps by a lot 
- `roaring` doesn't support uint64 bitmaps

index/segment/mem/postings_list.go
prateek: Sold. I'll remove the method

index/segment/mem/postings_list.go
prateek: `intersect` is definitely better than `andTo`. I don't like that it implies the underlying object `d *docIDs` is being changed. How would you feel about `project` or `intersectWith`

jeromefroe: I'd prefer `intersectWith` personally, `project` seems a little too vague.

doc/doc.go
jeromefroe: It seems more natural to me to take a slice of (key,value) pairs than a slice of keys and a slice of values. Perhaps we expose a `Field` type (I think there's one in the previous diff) and then we take a slice of them?

index/segment/mem/postings_list.go
prateek: sold, made the change

doc/doc.go
prateek: Sure, all depends on how the original data is stored. E.g. if a user has a `[]Field` type, they can construct a document without using this ctor; the document type and it's fields are all exported. 

Lets come back and address this once we know what input data looks like on the m3db side.

index/segment/mem/index.go
jeromefroe: I don't see how the fact that users don't have the internal ID would factor in, we would need to look up the ID anyway, and so would already have access to it. We will need to handle compactions for stored fields, but we can do that independently of actual merges so we can optimize when it makes the most sense to do it. In regards to the other concerns:

- I don't see how we would have contention given that we will likely only have a single open segment at a time. Even if we did have multiple open segments, getting a new ID is just an atomic increment, and should be orders of magnitude less expensive than actually inserting documents into the index.
- If we apply delta encoding, the memory footprint shouldn't actually increase.
- We can coerce the `uint64` to `uint32` to start if we want to use Roaring until we use a more specialized approach which natively supports `uin64`.

prateek: I don't follow. If we did Deletes/Updates by first retrieving the given document's internal id, and correspondingly deleting/updating the doc, then yeah, we'd get some benefit by having a doc id over the index lifecycle. 

But, we're going to do updates/deletes identical to insert. i.e. if we did and insert, and then a delete - we'd have 2 internal doc ids, each pointing to distinct documents that are identical until we compacted. The same thing for updates. i.e. we'd have to compact the fields for all documents. 

What am I missing?

jeromefroe: Might be easier to discuss in person and see where the ID per segment takes us to start. Like some of the other things, it's probably not worth getting hung up on this now lol.

index/segment/mem/postings_list.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// early terminate if we've expanded capacity since releasing lock last
+	if size = len(p.postings); int(t) < size {
+		p.Unlock()
+		return
+	}
+
+	// amortized O(1) expansion
+	factor := math.Ceil(float64(t) / float64(len(p.postings)))
+	if factor < 2 {
```
jeromefroe: Is it worth expanding the size of the slice linearly past some limit? For example, we always double up to X MB, at which point we expanding capacity X bytes at a time?

index/segment/mem/simple_terms_dict.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func (t *simpleTermsDictionary) Insert(field doc.Field, i segment.DocID) error {
+	fieldName := string(field.Name)
+	fieldValues := t.getOrAddFieldName(fieldName)
+	fieldValue := string(field.Value)
+	return fieldValues.addDocIDForValue(fieldValue, i, t.postingsList)
+}
+
+func (t *simpleTermsDictionary) Fetch(f segment.Filter) (docIDsSet, error) {
```
jeromefroe: Should we support multiple filters here so we can do things like: `service:foo AND env:production`?

index/segment/mem/types.go
```diff
@@ -62,3 +54,75 @@ type Options interface {
 	// NumInsertWorkers returns the number of insert workers.
 	NumInsertWorkers() int
 }
+
+// immutableDocIDsSet is a memory efficient collection of docIDs. It only permits
+// for read-only operations.
+type immutableDocIDsSet interface {
```
jeromefroe: Seems like this is what is traditionally called a postings list, whereas the `postingsList` interface below is a manager of different postings lists. Is it worth updating the terminology perhaps?

index/segment/mem/index.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (i *memIndex) insertDocument(doc document) error {
+	// insert document into master doc id -> doc map
+	i.documentsLock.Lock()
+	i.documents[doc.docID] = doc
+	i.documentsLock.Unlock()
+
+	// insert each of the indexed fields into the reverse index
+	for _, field := range doc.Fields {
```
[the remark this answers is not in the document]
prateek: evaluate perf impact and if it's shit, put a note at the least about this.

index/segment/mem/index.go
```diff
@@ -0,0 +1,172 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		// i.e. we don't have any documents for the given filter, can early terminate entire fn
+		if fetchedIds == nil {
+			return nil, nil
+		}
+
+		if !docsSet {
```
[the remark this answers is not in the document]
prateek: sure

index/segment/mem/simple_terms_dict.go
prateek: `segment.Query()` supports the kind of conjunctions you're thinking of. The `Fetch()` on a terms dicts is per filter

index/segment/mem/types.go
prateek: Sure thing

index/segment/mem/index.go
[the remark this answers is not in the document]
prateek: put a note down

index/segment/mem/index.go
```diff
@@ -0,0 +1,172 @@
...
+func (i *memIndex) Query(query segment.Query) ([]doc.Document, error) {
+	// TODO(prateek): timeout/early termination once we know we're done
+
+	var (
+		docsSet         = false
+		candidateDocIds docIDsSet
+	)
+
+	for _, filter := range query.Filters {
+		filter := filter
```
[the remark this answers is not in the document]
prateek: extracted to new `Searcher` interface

index/segment/mem/index.go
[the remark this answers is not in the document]
prateek: done

index/segment/mem/index.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+
+		if !docsSet {
+			candidateDocIds = fetchedIds
+			docsSet = true
+			continue
+		}
+
+		// update candidate set
+		candidateDocIds.Intersect(fetchedIds)
```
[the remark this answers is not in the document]
prateek: put a note down to try out both methods

index/segment/mem/postings_list.go
prateek: meh, lets see when we look at allocation profiles

index/segment/mem/index.go
```diff
@@ -0,0 +1,150 @@
...
+	"sync"
+
+	"github.com/m3db/m3ninx/doc"
+	"github.com/m3db/m3ninx/index/segment"
+
+	"github.com/uber-go/atomic"
+)
+
+// TODO(prateek): investigate impact of native heap
+type memIndex struct {
```
jeromefroe: Should this be a `memSegment`?

index/segment/mem/types.go
```diff
@@ -50,15 +42,69 @@ type Options interface {
...
+		filterValue []byte,
+		isRegex bool,
+	) (
+		candidateDocIDs segment.PostingsList,
+		pendingFilterFn predicate,
+		err error,
+	)
+
+	// Fetch retrieves the list of documents in the given postings list.
+	Fetch(p segment.PostingsList) ([]doc.Document, error)
```
jeromefroe: It seems a little strange to me that this interface has methods to return both postings lists and documents. If we are guaranteeing that the set of documents contained in segments is disjoint then it seems we should just support a `Fetch` on segments method which takes in a query and returns the documents since the postings list would be a detail internal to the segment. Then the index which contains multiple segments would be responsible for merging those documents into a single result.

index/segment/postings_list.go
```diff
@@ -0,0 +1,132 @@
...
+	d.Unlock()
+}
+
+func (d *roaringPostingsList) Intersect(other ImmutablePostingsList) {
+	o, ok := other.(*roaringPostingsList)
+	if !ok {
+		panic("Intersect only supported between roaringDocId sets")
+	}
+
+	d.Lock()
```
jeromefroe: nit: perhaps reverse the order of the locks so that the critical section for the write lock is shorter than that of the read lock?

index/segment/postings_list_pool.go
```diff
@@ -0,0 +1,49 @@
...
+type postingsListPool struct {
+	pool pool.ObjectPool
+}
+
+// NewPostingsListPool returns a new PostingsListPool.
+func NewPostingsListPool(opts pool.ObjectPoolOptions) PostingsListPool {
+	p := &postingsListPool{
+		pool: pool.NewObjectPool(opts),
+	}
+	p.pool.Init(func() interface{} {
```
jeromefroe: Should we pass the init function as a parameter so we can change what type of postings list is returned?

index/segment/types.go
```diff
@@ -136,3 +105,57 @@ type Filter struct {
...
+}
+
+// PostingsList is a memory efficient collection of docIDs.
+type PostingsList interface {
+	ImmutablePostingsList
+
+	// Insert inserts the given id to set.
+	Insert(i DocID)
+
+	// Intersect modifies the receiver set to contain only those ids by both sets.
```
jeromefroe: Would it be possible to move these methods out of the interface (i.e. `Intersect(pLists ImmutablePostingsList...) ImmutablePostingsList`)? Requiring it to mutate the underlying postings list might not make sense for all postings list implementations, and it will guarantee that we have to make a copy of at least one of the postings lists when responding to a query, but if we intersect them all together we can may be able to avoid any copy or allocate much less memory when we do.

index/segment/mem/index.go
prateek: 👍 Will do

index/segment/mem/simple_terms_dict.go
```diff
@@ -0,0 +1,206 @@
...
+import (
+	"regexp"
+	"sync"
+
+	"github.com/m3db/m3ninx/doc"
+	"github.com/m3db/m3ninx/index/segment"
+)
+
+// simpleTermsDictionary uses two-level map to model a terms dictionary.
+// i.e. fieldName -> fieldValue -> postingsManagerOffset
```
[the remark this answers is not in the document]
prateek: 👍 will do

index/segment/mem/types.go
```diff
@@ -50,15 +42,69 @@ type Options interface {
...
+// predicate returns a bool indicating if the document matched the
+// provided criterion, or not.
+type predicate func(d doc.Document) bool
+
+// queryable is the base contract required for any mem segement implementation to be used by a `searcher`.
+type queryable interface {
+	// Filter retrieves the PostingsList for the filter, and any additional
+	// filtering criterion that must be applied in post processing.
+	Filter(
+		filterName []byte,
```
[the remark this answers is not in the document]
prateek: Sure

index/segment/mem/types.go
prateek: These methods are used by a `searcher` which is internal to a segment, these methods can't be used by external consumers. The abstraction afforded to external users is the one you're suggesting ~ `segment.Query(q Query)`

index/segment/postings_list.go
prateek: sure

index/segment/postings_list_pool.go
prateek: Sure

index/segment/mem/index.go
prateek: done

index/segment/mem/simple_terms_dict.go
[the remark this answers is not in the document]
prateek: done

index/segment/mem/types.go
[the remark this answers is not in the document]
prateek: done

index/segment/postings_list.go
prateek: done

index/segment/postings_list_pool.go
prateek: done

index/segment/types.go
prateek: Lets defer this refactor till we have an additional PostingsList implementation/need to do this for other reasons (perf/allocation/...).

index/segment/mem/types.go
jeromefroe: Oh, I see now, `Filter` gets the postings list for the individual terms, then the `searcher` does what it needs to based on the query (intersect for And, union for OR, etc.) and then retrieves the actual documents with `Fetch`?

prateek: Yep exactly.

index/segment/mem/searcher.go
```diff
@@ -74,5 +111,62 @@ func (s *sequentialSearcher) Query(query segment.Query) ([]doc.Document, error)
 	}
 
 	// TODO: once we support multiple segments, we'll have to merge results
-	return s.queryable.Fetch(candidateDocIds)
+	return s.queryable.Fetch(candidateDocIds, matchPredicates(predicates).Fn())
```
jeromefroe: Do we have to execute any subqueries in `Query` as well?

index/segment/mem/searcher.go
```diff
@@ -21,37 +21,70 @@
...
 )
 
+var (
+	errFirstFilterMustNotBeNegation = fmt.Errorf("first filter must be non-negation")
+)
+
+type negationMergeFn func(x segment.PostingsList, y segment.ImmutablePostingsList)
+
+func differenceNegationFn(x segment.PostingsList, y segment.ImmutablePostingsList) { x.Difference(y) }
+func unionNegationFn(x segment.PostingsList, y segment.ImmutablePostingsList)      { x.Union(y) }
```
jeromefroe: When would this be used? My initial reaction is that we would only ever want to take the difference when we get a negation.

index/segment/mem/searcher.go
```diff
@@ -74,5 +111,62 @@ func (s *sequentialSearcher) Query(query segment.Query) ([]doc.Document, error)
...
+type orderFiltersByNonNegated []segment.Filter
+
+func (of orderFiltersByNonNegated) Len() int           { return len(of) }
+func (of orderFiltersByNonNegated) Swap(i, j int)      { of[i], of[j] = of[j], of[i] }
+func (of orderFiltersByNonNegated) Less(i, j int) bool { return !of[i].Negate && of[j].Negate }
+
+// validate any assumptions we have about queries.
+func validateQuery(q segment.Query) error {
+	// assuming we only support AndConjuctions for now.
+	if q.Conjunction != segment.AndConjunction {
```
jeromefroe: Not an issue now, but is it worth calling out that we probably won't support union with negation filters? i.e. queries such as `key1=value1 OR key2!=value2` will be illegal since it's a superset of `key2!=value2` which is illegal

index/segment/mem/segment_test.go
```diff
@@ -0,0 +1,284 @@
...
+		found := false
+		for _, ret := range docs {
+			if string(req) == string(ret.ID) {
+				found = true
+			}
+		}
+		t.True(found, string(req))
+	}
+}
+
```
jeromefroe: Maybe add tests for subqueries also?

doc/doc.go
robskillington: Maybe we can just make this `Fields` and then users can wrap it up however they want?

i.e.
```
type Fields interface {
  Iterator() FieldsIterator
}
```

doc/types.go
```diff
@@ -20,6 +20,15 @@
 
 package doc
 
+// ReservedFieldName signifies the field names which are reserved for internal
+// use.
+type ReservedFieldName string
```
robskillington: Can we get away with needing to used reserved fields?  Can't we simply store properties somewhere else than the fields themselves if we have special need for them?

index/segment/mem/results_iter.go
```diff
@@ -0,0 +1,129 @@
...
+	if !ok {
+		r.release()
+		r.hasNext = false
+		r.done = true
+		return
+	}
+	// setup values so that next time Next() is called, we can rotate
+	// next -> current
+	r.next = d
+	r.hasNext = ok
```
robskillington: This will always be true no?

index/segment/mem/results_iter.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (r *resultsIter) release() {
+	// need this cast because we're provided an immutable postings lists
+	// and we pool mutable postings lists
+	pl, ok := r.postingsList.(segment.PostingsList)
+	if ok {
+		r.queryable.Options().PostingsListPool().Put(pl)
```
robskillington: Might be worth also nil'ing out our ref so there's no potential for using it after its placed back in the pool.  `r.postingList = nil`

index/segment/mem/results_iter.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+		// ensure document matches predicateFn
+		if r.predicateFn(doc.Document) {
+			return doc, true
+		}
+	}
+}
+
+func (r *resultsIter) Next() bool {
```
robskillington: Need to support the following flow (not sure this is currently supported?):
```go
iter := thing.Query()
for iter.Next() { // note: Next() is called before consuming the first entry
  v, t := iter.Current()
  // use
}
```

index/segment/mem/results_iter.go
```diff
@@ -0,0 +1,129 @@
...
+			return document{}, false
+		}
+
+		id := r.idsIter.Current()
+		doc, ok := r.queryable.FetchDocument(id)
+		if !ok {
+			// i.e. we were given a reference to an illegal document ID by the
+			// postings list iter. This should never happen, but listing here for
+			// completeness.
+			// TODO: log error/metric for ^^
```
robskillington: Our current pattern is to add an `err` field to the iterator and an `Err() error` method on the iterator.

Then at the end you check if there was an error during iteration:

index/segment/mem/searcher.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+
+		// early terminate if we don't have any docs in candidate set
+		if candidateDocIds.IsEmpty() {
+			return nil, nil, nil
+		}
+	}
+
+	// TODO: once we support multiple segments, we'll have to merge results
+	return candidateDocIds, matchPredicates(predicates).Fn(), nil
```
robskillington: Instead of creating a heap allocated slice and a heap allocated fn, wouldn't it be better to return just a single wrapper type that is stack allocated?

i.e.

index/segment/mem/searcher.go
```diff
@@ -0,0 +1,176 @@
...
+// validate any assumptions we have about queries.
+func validateQuery(q segment.Query) error {
+	// assuming we only support AndConjuctions for now.
+	if q.Conjunction != segment.AndConjunction {
+		return fmt.Errorf("query: %v has an invalid conjuction: %v", q, q.Conjunction)
+	}
+
+	// ensure query level have at-least one filter or sub-query
+	if len(q.Filters) == 0 && len(q.SubQueries) == 0 {
+		return fmt.Errorf("empty query specified")
```
robskillington: nit: Declare err with a `errNoQuery` instead of fmt.Errorf since its a constant string.

index/segment/mem/segment_simple.go
```diff
@@ -0,0 +1,133 @@
...
+
+// TODO(prateek): investigate impact of native heap
+type simpleSegment struct {
+	opts     Options
+	id       segment.ID
+	docIDGen *atomic.Uint32
+
+	// internal docID -> document
+	documentsLock sync.RWMutex
+	documents     map[segment.DocID]document // TODO(prateek): measure perf impact of slice v map here
```
robskillington: When I find myself prefixing fields the same I find it best to make it composite.

For instance.

index/segment/types.go
```diff
@@ -44,66 +35,35 @@ type Segment interface {
...
-	Open() error
-
-	// Close closes the reader.
-	Close() error
-}
-
-// DocID is document identifier internal to each index segment. It's limited to
+// DocID is a document identifier internal to each index segment. It's limited to
 // be at most 2^32.
 type DocID uint32
```
robskillington: Any reason not to use int64?  4bill seems too small, especially since we already have 4bill in our current index (granted that is globally not per zone).

index/segment/mem/segment_simple.go
```diff
@@ -0,0 +1,133 @@
...
+	return i.insertDocument(document{
+		Document: d,
+		docID:    segment.DocID(i.docIDGen.Inc()),
+	})
+}
+
+func (i *simpleSegment) insertDocument(doc document) error {
+	// insert document into master doc id -> doc map
+	i.documentsLock.Lock()
+	i.documents[doc.docID] = doc
```
robskillington: We'll have a segment per shard yeah?  Otherwise a simple top level mutex seems like it might cause a lot of contention.

index/segment/mem/types.go
```diff
@@ -50,15 +42,48 @@ type Options interface {
...
+// termDictionary represents a mapping from doc.Field -> doc
+type termsDictionary interface {
+	// Insert inserts a mapping from field `f` to document ID `docID`.
+	Insert(f doc.Field, docID segment.DocID) error
 
-	// SetNumInsertWorkers sets the number of insert workers.
-	SetNumInsertWorkers(value int) Options
+	// Fetch returns all the docIDs matching the given filter.
+	// NB(prateek): the returned PostingsList is safe to modify.
+	Fetch(fieldName []byte, fieldValueFilter []byte, isRegexp bool) (segment.PostingsList, error)
```
robskillington: Might be worth making the `isRegexp` field her an options struct.  I see there being more of these type of options in the future and might make cross-repo changes hard if its not a struct.

index/segment/mem/simple_terms_dict.go
```diff
@@ -0,0 +1,198 @@
...
+	return &simpleTermsDictionary{
+		fieldNames: make(map[string]*fieldValuesMap, opts.InitialCapacity()),
+		opts:       opts,
+	}
+}
+
+func (t *simpleTermsDictionary) Insert(field doc.Field, i segment.DocID) error {
+	fieldName := string(field.Name)
+	fieldValues := t.getOrAddFieldName(fieldName)
+	fieldValue := string(field.Value)
```
robskillington: Any possibility of avoiding all these string casts?

index/segment/mem/simple_terms_dict.go
```diff
@@ -0,0 +1,198 @@
...
+
+	"github.com/m3db/m3ninx/doc"
+	"github.com/m3db/m3ninx/index/segment"
+)
+
+// simpleTermsDictionary uses two-level map to model a terms dictionary.
+// i.e. fieldName -> fieldValue -> postingsList
+type simpleTermsDictionary struct {
+	fieldNamesLock sync.RWMutex
+	fieldNames     map[string]*fieldValuesMap
```
robskillington: Any possibility avoiding this using strings for map lookup?

index/segment/mem/simple_terms_dict.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (t *simpleTermsDictionary) Fetch(
+	fieldName []byte,
+	fieldValueFilter []byte,
+	isRegexp bool,
+) (segment.PostingsList, error) {
+	// check if we know about the field name
+	t.fieldNamesLock.RLock()
+	fieldValues, ok := t.fieldNames[string(fieldName)]
```
robskillington: Seems expensive to cast to string for each fetch?

index/segment/mem/simple_terms_dict.go
```diff
@@ -0,0 +1,198 @@
...
+	// cheap read lock to see if it already exists
+	t.fieldNamesLock.RLock()
+	fieldValues, ok := t.fieldNames[fieldName]
+	t.fieldNamesLock.RUnlock()
+	if ok {
+		return fieldValues
+	}
+
+	// acquire write lock and create
+	t.fieldNamesLock.Lock()
```
robskillington: Probably cheap enough to `defer ...Unlock()` here.

index/segment/mem/simple_terms_dict.go
robskillington: Also similar to the `documentsLock` and `documents` comment above.

index/segment/mem/simple_terms_dict.go
```diff
@@ -0,0 +1,198 @@
...
+	return fieldValues
+}
+
+type fieldValuesMap struct {
+	sync.RWMutex
+
+	opts Options
+
+	// fieldValue -> postingsList
+	values map[string]segment.PostingsList
```
robskillington: Similar comment re this map string lookup.

index/segment/mem/simple_terms_dict.go
```diff
@@ -0,0 +1,198 @@
...
+	f.RUnlock()
+
+	// we have a postingsList, mark the docID and move on.
+	if ok {
+		pid.Insert(i)
+		return nil
+	}
+
+	// postingsList doesn't exist, time to acquire write lock
+	f.Lock()
```
robskillington: Seems cheap enough here to `defer ...Unlock()`.

index/segment/mem/simple_terms_dict.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func newRegexPredicate(valueFilter []byte) (valuePredicate, error) {
+	filter := string(valueFilter)
+	re, err := regexp.Compile(filter)
+	if err != nil {
+		return nil, err
+	}
+
+	return re.MatchString, nil
```
robskillington: So we're not using trigrams for v1?

index/segment/postings_list.go
```diff
@@ -0,0 +1,143 @@
...
+		i: d.bitmap.Iterator(),
+	}
+}
+
+func (d *roaringPostingsList) Clone() PostingsList {
+	d.RLock()
+	// TODO: It's cheaper to Clone than to cache roaring bitmaps, see
+	// `postings_list_bench_test.go`. Their internals don't allow for
+	// pooling at the moment. We should address this when get a chance
+	// (move to another implementation / address deficiencies).
```
robskillington: Ugh (ha)

index/segment/mem/results_iter.go
prateek: Sure, I'll make it explicit.

index/segment/mem/results_iter.go
prateek: Sure thing

index/segment/mem/results_iter.go
prateek: Yep, so the implementation addresses that by keeping vars for both current and next values. On init, we only set the next value. Every call to `Next()` rotates next -> current, and fetches the next (if available)

index/segment/mem/searcher.go
prateek: sure thing

index/segment/mem/segment_simple.go
prateek: Sure thing.

index/segment/mem/simple_terms_dict.go
prateek: sure thing

index/segment/mem/simple_terms_dict.go
prateek: We will, #7 is already in the pipeline :) Holding off adding that here to keep PRs small.

index/segment/mem/segment_simple.go
prateek: Nah I'm thinking of keeping segment and shard independent. It'll be too expensive to keep a segment per shard on the read side – every fetch would have to hit every segment in each shard; i.e. thousands of map lookups. 

My plan for contention was to have an insert channel at the top-level index.

doc/types.go
jeromefroe: Perhaps we should update the comment, because these fields aren't really internal fields. The user can, and is in fact expected, to use the ID field as well. It's just that this field has a significance attached to it (each ID field uniquely determines a document) that other fields do not have.

index/segment/mem/segment_simple.go
jeromefroe: If the indices are independent of the shards though how do we plan on splitting indices when nodes join a cluster and shards are assigned to different hosts? We won't be able to just move the index along with the shard like we would be able to do with an index per shard approach.

index/segment/types.go
jeromefroe: We picked `int32` because there's much greater support for it in algorithms for bitmap indices (Roaring) and posting lists compression (Frame of Reference and Simple Family) than `int64`. Not that they couldn't be adopted for `int64` just that it would require more up-front work on our end (for example, we can use the go implementation of Roaring as is). We still haven't settled on this though. I think when we last left off I was in favor of an ID per index, with `int64` if we felt we would have indices with more than 4 billion documents, and Prateek was in favor of an ID per segment, with `int32` since it's how to imagine a single segment containing more than 4 billion IDs, and we were going to revisit this issue later on when we had more benchmarks. Which is all to say I think this is still largely up in the air lol

doc/types.go
prateek: Yeah this isn't used/needed atm. Will remove for now.

index/segment/mem/results_iter.go
prateek: Good call, will add.

index/segment/mem/segment_simple.go
prateek: Two parts: 
- For the node with the leaving shard, we'd expect the shard to issue a `Delete()` for all it's IDs so that the index doesn't have any zombie docs
- For the node with the joining shard, it'd bootstrap the same as the normal process after bootstrap and get the appropriate Insert() called.

index/segment/types.go
prateek: Pretty much what Jerome said. Only thing to add is the DocID is internal to each Segment, which are per node. I don't think we'll ever hit a single node with 4B documents so this isn't a size restriction.

index/segment/mem/segment_simple.go
prateek: ended up using anonymous structs for this:

```
type simpleSegment struct {
  // ...
  docs struct {
    sync.RWMutex
    values map[segment.DocID]document
  }
}
```

index/segment/mem/simple_terms_dict.go
prateek: Yeah it'll require we handroll a hash map implementation unfortunately. I opened #11 for this, and will leave TODOs referencing it.

index/segment/mem/simple_terms_dict.go
prateek: same as above, left TODO at the top level struct about #11

index/segment/mem/simple_terms_dict.go
prateek: Yeah it's suboptimal for sure. I think it's the better interface to still take a `[]byte` as input cause that's probably where we want to end up. We can avoid this casts once we roll our own hashmap.

index/segment/mem/simple_terms_dict.go
prateek: Yeah just think it's a small enough function that it doesn't hurt readability to not use a defer here. Only one exit point in the if right after this needs the extra unlock call.

index/segment/mem/simple_terms_dict.go
prateek: yep, left a todo here too

index/segment/mem/simple_terms_dict.go
prateek: same as above, will change if you feel strongly.

doc/doc.go
robskillington: As discussed perhaps just remove the constructor for now?

prateek: Yep, removed.

index/segment/mem/segment_simple.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/m3db/m3ninx/index/segment"
+
+	"github.com/uber-go/atomic"
+)
+
+// TODO(prateek): investigate impact of native heap
+type simpleSegment struct {
+	opts     Options
+	id       segment.ID
+	docIDGen *atomic.Uint32
```
robskillington: So these doc IDs are only valid within a segment yeah?  Just curious.

index/segment/mem/results_iter.go
```diff
@@ -0,0 +1,144 @@
...
+func (r *resultsIter) getNextDocument() (document, bool, error) {
+	// need to loop as we may filter out values due to the predicate function
+	for {
+		hasNext := r.idsIter.Next()
+		if !hasNext {
+			return document{}, false, nil
+		}
+
+		id := r.idsIter.Current()
+		doc, ok := r.queryable.FetchDocument(id)
```
robskillington: Feels weird that we actually iterate but then do a hash lookup per each document we iterate over? (when using a simple segment)

index/segment/mem/segment_simple.go
```diff
@@ -0,0 +1,133 @@
...
+	docs, err := i.termsDict.Fetch(f.FieldName, f.FieldValueFilter, f.Regexp)
+	return docs, nil, err
+}
+
+func (i *simpleSegment) Delete(d doc.Document) error {
+	panic("not implemented")
+}
+
+func (i *simpleSegment) Size() uint32 {
+	panic("not implemented")
```
robskillington: We can't return just len(values)?

index/segment/mem/segment_simple.go
```diff
@@ -0,0 +1,133 @@
...
+) (segment.PostingsList, matchPredicate, error) {
+	docs, err := i.termsDict.Fetch(f.FieldName, f.FieldValueFilter, f.Regexp)
+	return docs, nil, err
+}
+
+func (i *simpleSegment) Delete(d doc.Document) error {
+	panic("not implemented")
+}
+
+func (i *simpleSegment) Size() uint32 {
```
robskillington: Same question about segment size and 4bill here, i.e. any reason not to just use uint64?

index/segment/mem/simple_terms_dict.go
robskillington: Ok np, sounds good.

index/segment/mem/simple_terms_dict.go
robskillington: Ok np, sounds good.

index/segment/mem/simple_terms_dict.go
```diff
@@ -0,0 +1,206 @@
...
+	if !regexp {
+		return f.fetchExact(valueFilter), nil
+	}
+
+	// otherwise, we have to iterate over all known values
+	pred, err := newRegexPredicate(valueFilter)
+	if err != nil {
+		return nil, err
+	}
+	var postingsLists []segment.PostingsList
```
robskillington: Probably worth allocating some initial size, maybe a fixed factor * len(f.values)?

index/segment/mem/simple_terms_dict.go
robskillington: Makes sense, ok cool.

index/segment/mem/simple_terms_dict.go
```diff
@@ -0,0 +1,206 @@
...
+	filter := string(valueFilter)
+	re, err := regexp.Compile(filter)
+	if err != nil {
+		return nil, err
+	}
+
+	return re.MatchString, nil
+}
+
+type valuePredicate func(v string) bool
```
robskillington: Maybe move this to the top of this file?

index/segment/mem/types.go
robskillington: Thoughts on this?

jeromefroe: +1 for the options struct since we may want to distinguish glob expressions from regexes at some point and who knows what else

index/segment/mem/segment_simple.go
prateek: Yep this field is used to generate the next segment.DocID, internal to each segment. 

A single node should never be responsible for that order of magnitude (Billions) of documents. Opened #12 to track longer term.

index/segment/mem/segment_simple.go
prateek: Per above, tracking in #12

index/segment/mem/simple_terms_dict.go
prateek: sure thing

index/segment/mem/simple_terms_dict.go
prateek: sure thing

index/segment/mem/segment_simple.go
prateek: Yep will do

index/segment/mem/types.go
prateek: Sorry missed this earlier. Sounds good, will do.

index/segment/mem/results_iter.go
prateek: I have a TODO in there to see impact of making `FetchDocument` slice backed instead of hash backed. Will that might help address the concern?

The main reason for the is it lets us do the intersection cheaply across filters by using a 
postings list of the ints. Probably will retain a `filter -> postingsList` cache in the near term that'll need this too.

index/segment/mem/types.go
prateek: done

index/segment/mem/segment_simple.go
prateek: done

index/segment/mem/simple_terms_dict.go
prateek: done

index/segment/mem/simple_terms_dict.go
prateek: done

index/segment/mem/results_iter.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (r *resultsIter) Next() bool {
+	// ensure internal state is valid
+	if r.done || r.err != nil {
+		return false
+	}
+
+	// early terminate if we do not have another value queued
+	if !r.hasNext {
+		return false
```
robskillington: nit: combine the two `if ... { return false }` and use named bool's perhaps if it seems confusing?  (then it's also self documenting)

i.e.
```go
valid := !r.done && r.err == nil
remaining := r.hasNext
if !valid || !remaining {
  return false
}
```

prateek: sure thing.

index/segment/mem/segment_simple.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	docs := make([]document, 2*(size+1))
+	copy(docs, i.docs.values)
+	i.docs.values = docs
+	i.docs.Unlock()
+}
+
+func (i *simpleSegment) insertDocument(doc document) error {
+	i.ensureSufficientCapacity(int64(doc.docID))
```
robskillington: Combine the RLocks together perhaps so just need to acquire once?

index/segment/mem/searcher.go
```diff
@@ -0,0 +1,178 @@
...
+	pendingFilterFn matchPredicate,
+	err error,
+) {
+	if err := validateQuery(query); err != nil {
+		return nil, nil, err
+	}
+
+	// order filters to ensure the first filter has no-negation
+	filters := orderFiltersByNonNegated(query.Filters)
+	sort.Sort(filters)
```
robskillington: This is race safe yeah?  Just pondering if anyone else holds refs to query, etc and if in the future there will be parallel searches, etc.

index/segment/mem/simple_terms_dict.go
```diff
@@ -0,0 +1,212 @@
...
+	pred, err := newRegexPredicate(valueFilter)
+	if err != nil {
+		return nil, err
+	}
+
+	f.RLock()
+	initLength := int(regexpPostingsListInitFactor * float64(len(f.values)))
+	postingsLists := make([]segment.PostingsList, 0, initLength)
+	for value, list := range f.values {
+		if pred(value) {
```
robskillington: At some point we might want to evaluate the predicate outside of needing to hold a lock, but that's something we can attack when we run into it I suppose (contending for this lock that is).

index/segment/mem/simple_terms_dict.go
```diff
@@ -0,0 +1,212 @@
...
+
+	// create new posting id for the term, and insert into fieldValues
+	pid = f.opts.PostingsListPool().Get()
+	f.values[value] = pid
+	f.Unlock()
+	pid.Insert(i)
+	return nil
+}
+
+func (f *fieldValuesMap) fetchLists(valueFilter []byte, regexp bool) ([]segment.PostingsList, error) {
```
robskillington: At some point we might like to turn this into an iterator, to avoid the array alloc for returning the result, etc.  Is tough when using a map backed structure for the field values map however.   

Just thought I'd mention it, this is good for now though.

index/segment/postings_list.go
```diff
@@ -0,0 +1,143 @@
...
+func (d *roaringPostingsList) Insert(i DocID) {
+	d.Lock()
+	d.bitmap.Add(uint32(i))
+	d.Unlock()
+}
+
+func (d *roaringPostingsList) Intersect(other ImmutablePostingsList) {
+	o, ok := other.(*roaringPostingsList)
+	if !ok {
+		panic("Intersect only supported between roaringDocId sets")
```
robskillington: Hm I'd be in preference of bubbling this all up as errors tbh.  I know this is an application bug that ideally would be compile time enforced but because of the interfaces used here cannot be - having said that I could definitely imagine in the future needing to return an error for intersect/difference/etc so it'd probably better to just add error as a type to these interface methods and bubble up the errors instead.

Otherwise we'll start to use this as a pattern in other places or take it as "acceptable."  

Everything we write should try as much as possible to avoid causing panics themselves and bubble up errors so callers can make the call on whether an operation causing an error is fatal enough to panic.

index/segment/postings_list_pool.go
```diff
@@ -0,0 +1,55 @@
...
+// NewPostingsListPool returns a new PostingsListPool.
+func NewPostingsListPool(
+	opts pool.ObjectPoolOptions,
+	allocator PostingsListAllocateFn,
+) PostingsListPool {
+	p := &postingsListPool{
+		pool: pool.NewObjectPool(opts),
+	}
+	p.pool.Init(func() interface{} {
+		return allocator()
```
robskillington: Do you want to `pl.Reset()` before returning or should new alloc always have same state as a reset postingslist?

index/segment/types.go
```diff
@@ -136,3 +99,60 @@ type Filter struct {
...
+
+	// Iter returns an iterator over the known values.
+	Iter() PostingsIter
+
+	// Clone returns a copy of the set.
+	Clone() PostingsList
+}
+
+// PostingsList is a memory efficient collection of docIDs.
+type PostingsList interface {
```
robskillington: I think all the mutable operations here should probably return `error`.

index/segment/mem/searcher.go
prateek: yep it's only used in this one path (no other references). We'll eliminate the sort if we parallelise in the future.

index/segment/mem/simple_terms_dict.go
prateek: Yep fair. i'll leave a TODO about inspecting lock contention

index/segment/mem/simple_terms_dict.go
prateek: Yep sounds good.

index/segment/postings_list.go
prateek: Fair. I do think panics have a place (e.g. before allowing corrupt data to be written to disk), this isn't one of those cases. Will bubble error everywhere.

index/segment/postings_list_pool.go
prateek: yep, assuming the allocator returns a usable postings list to start. We reset on Put to ensure the objects in the pool are usable.

index/segment/types.go
prateek: Sure thing.

index/segment/mem/results_iter.go
prateek: done

index/segment/mem/segment_simple.go
prateek: done

index/segment/mem/searcher.go
prateek: done

index/segment/mem/simple_terms_dict.go
prateek: done

index/segment/mem/simple_terms_dict.go
prateek: done.

index/segment/postings_list.go
prateek: done.

index/segment/types.go
prateek: done

index/segment/mem/results_iter.go
```diff
@@ -0,0 +1,141 @@
...
+		r.err = err
+		r.done = true
+		return
+	}
+
+	if !ok {
+		r.release()
+		r.hasNext = false
+		r.done = true
+		return
```
robskillington: nit: You can combine the two together, don't need to if you don't want to though.

index/segment/mem/searcher.go
robskillington: Ok, np - maybe worth perhaps adding a comment w.r.t to be careful about modifying it if the parallelization is done in the future.

index/segment/mem/segment_simple.go
```diff
@@ -0,0 +1,171 @@
...
+
+func (i *simpleSegment) FetchDocument(id segment.DocID) (document, error) {
+	if uint32(id) > i.Size() {
+		return document{}, errUnknownDocID
+	}
+
+	i.docs.RLock()
+	d := i.docs.values[id]
+	i.docs.RUnlock()
+	return d, nil
```
robskillington: Hm so I believe this method is a little racey, the `docIDGen` is incremented before the slice is grown potentially.  Wouldn't it just be safer to grab the RLock() and check the length of the slice?  You have to grab the RLock() in the case that its valid anyway.

index/segment/mem/simple_terms_dict.go
```diff
@@ -0,0 +1,213 @@
...
+	// create new posting id for the term, and insert into fieldValues
+	pid = f.opts.PostingsListPool().Get()
+	f.values[value] = pid
+	f.Unlock()
+	return pid.Insert(i)
+}
+
+// TODO: consider returning an iterator here, this would require some kind of ordering semantics
+// on the underlying map tho.
+func (f *fieldValuesMap) fetchLists(valueFilter []byte, regexp bool) ([]segment.PostingsList, error) {
```
robskillington: Any reason to use `regexp bool` instead of `opts termFetchOptions` here?

index/segment/mem/simple_terms_dict.go
```diff
@@ -0,0 +1,213 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	f.RLock()
+	initLength := int(regexpPostingsListInitFactor * float64(len(f.values)))
+	postingsLists := make([]segment.PostingsList, 0, initLength)
+	for value, list := range f.values {
+		// TODO: evaluate lock contention caused by holding on to the read lock
+		// while evaluating this predicate.
```
robskillington: You can do what we do in the M3DB shard lookup, keep both a list and a map, then iterate over the list in a batched fashion to avoid holding lock for an entire scan and use lock in batches of work.

index/segment/postings_list_pool.go
robskillington: Usually consolidating the "clean state" to a single method is ideal.

index/segment/mem/searcher.go
prateek: nah you're right, i'll make an explicit copy of the query object to avoid this issue.

index/segment/mem/results_iter.go
prateek: fair, will do

index/segment/mem/segment_simple.go
prateek: Yeah, good catch. Thanks!

index/segment/mem/simple_terms_dict.go
prateek: Yep, for sure. The hope is the map backed segment never gets large enough for this to be an issue, will address if it does.

index/segment/mem/simple_terms_dict.go
prateek: Yeah no reason not to use the opts. Will make the change.

index/segment/mem/simple_terms_dict.go
prateek: done

index/segment/mem/simple_terms_dict.go
prateek: done

index/segment/mem/segment_simple.go
prateek: done

index/segment/mem/results_iter.go
prateek: done

index/segment/mem/searcher.go
prateek: done

PR #13, opened 2018-01-30
[mutable segment: part 1] Add initial implementation of trigram terms dictionary

jeromefroe:
This PR adds an implementation of a trigram-based dictionary based on #7 

cc @prateek


glide.yaml
```diff
@@ -8,6 +8,8 @@ import:
   - instrument
 - package: github.com/uber-go/atomic
   version: ^1.2.0
+- package: github.com/google/codesearch
```
prateek: could you use the m3db forked one, `github.com/m3db/codesearch`

jeromefroe: sure, wasn't aware we had a fork

index/segment/mem/trigram_terms_dict.go
```diff
@@ -0,0 +1,267 @@
...
+
+	"github.com/m3db/m3ninx/doc"
+	"github.com/m3db/m3ninx/index/segment"
+
+	"github.com/google/codesearch/index"
+)
+
+var (
+	// sentinelTrigram is stored for every field so that we can perform match all lookups.
+	sentinelTrigram = ""
```
prateek: Would it be better to not do this? The main concern is the postings list corresponding to this entry will be pretty huge, which'd make the candidate search space for some queries large too. How do you feel about instead having the trigram dictionary return its unable to satisfy any queries needing this?

index/segment/mem/trigram_terms_dict.go
```diff
@@ -0,0 +1,267 @@
...
+//   (2) For each trigram `t` created in step 1, store the entry (value: `t`, docID: `i`) in the
+//       postings list for the field name "foo".
+//
+
+// trigramsTermsDict uses trigrams in a two-level map to model a terms dictionary.
+// i.e. fieldName > trigram(fieldValue) -> postingsList
+type trigramTermsDictionary struct {
+	opts Options
+
+	fields struct {
```
prateek: I don't follow the thought process here:
- would be more efficient to return the candidate docIDs from `Fetch` - that way, we can merge across multiple fields before performing validation.
- document -> fields is many one, should the map value be `[]doc.Field`?

index/segment/mem/trigram_terms_dict.go
jeromefroe: I'm not sure. My hesitation about the latter approach is that we break the abstraction provided by the segment interface. Namely, the index would have to know what kind of queries the segment supports. For example, if I have the term `foo:c1`, the value isn't long enough to produce any trigrams and would effectively be "invisible" in the index. Thus, even if there are very few terms with key `foo`, so that the postings list for the key is small, we would still have to return an error code indicating that we are unable to satisfy the query. If we want to still support searching for such terms then we could have a mechanism in the index to account for them but then we are exposing a lot of the implementation details of the trigram index. My inclination is to go with this admittedly more heavyweight approach so that we can cleanly satisfy the segment interface and reevaluate when we have more benchmarks.

index/segment/mem/trigram_terms_dict.go
jeromefroe: For the second point, yea, it should be `[]doc.Field`, I was mistaken because the `Insert` method only takes a single field, will update.

For the first point, if we just return the candidate docIDs from `Fetch` then we can return false positives. I wanted to avoid this since then the index would always have to check for false positives in the docIDs returned from a segment, even though some segments don't return any false positives.

index/segment/mem/trigram_terms_dict.go
```diff
@@ -0,0 +1,216 @@
...
+	"regexp/syntax"
+
+	"github.com/m3db/m3ninx/doc"
+	"github.com/m3db/m3ninx/index/segment"
+
+	"github.com/m3db/codesearch/index"
+)
+
+var (
+	errUnsupportedQuery = errors.New("query is not supported by the trigram terms dictionary")
```
prateek: suggestion(take/leave): clearer to make this exported if we're going to use it across files, even if it's only package local. e.g. when we have a segment using both trigram+simple, it'll need to be able to tell why the trigram fetch failed.

index/segment/mem/trigram_terms_dict.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	return candidateIDs, nil
+}
+
+func (t *trigramTermsDictionary) docIDsForTrigram(
+	fieldName []byte,
+	tri string,
+) (segment.ImmutablePostingsList, error) {
+	return t.backingDict.Fetch(fieldName, []byte(tri), termFetchOptions{isRegexp: false})
```
prateek: Would it be better to expose a `FetchString()` on the simpleTerms dictionary to avoid the []byte alloc here?

index/segment/mem/trigram_terms_dict.go
jeromefroe: I think my personal preference would be to leave it unexported to reduce the surface area of the package API. We can of course always change that in the future.

prateek: Sounds good

index/segment/mem/trigram_terms_dict.go
jeromefroe: That's a fair point. I'll take this as a TODO since I think we may want to evaluate our use of strings vs. slices throughout the package because we do something unnecessary conversions in other places as well.

PR #16, opened 2018-02-13
[mutable segment: part 3] Add benchmarks for terms dictionary implementations

jeromefroe:
Add benchmarks for terms dictionary implementations. The results are below:

```
$ go test -run NONE -bench Benchmark
goos: darwin
goarch: amd64
pkg: github.com/m3db/m3ninx/index/segment/mem
BenchmarkInsert_SimpleTermsDictionary-4              500           3178559 ns/op
BenchmarkInsert_TrigramTermsDictionary-4              50          26942359 ns/op
BenchmarkFetch_SimpleTermsDictionary-4               100          10602854 ns/op
BenchmarkFetch_TrigramTermsDictionary-4                3         365077453 ns/op
BenchmarkFetchRegex_SimpleTermsDictionary-4        20000            113381 ns/op
BenchmarkFetchRegex_TrigramTermsDictionary-4       30000             49800 ns/op
```

These come with the big caveat that when testing the `Fetch` performance of the trigram terms dictionary I didn't do any post filtering to remove any false positives. Since for most cases this will likely be necessary we may want to consider adding such filtering.

cc @prateek


index/segment/mem/terms_dict_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+import (
+	"bufio"
+	"encoding/json"
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/m3db/m3ninx/index/segment"
+
```
prateek: nit: newline

index/segment/mem/terms_dict_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/m3db/m3ninx/doc"
+)
+
+var (
+	benchDocs []doc.Document
+)
+
+func init() {
+	var err error
+	if benchDocs, err = readDocuments("../../../testdata/node_exporter.json", 2000); err != nil {
```
prateek: probably better to use the absolute path 

```
path := os.ExpandEnv("${GOPATH}/github.com/m3db/m3ninx/testdata/node_exporter.json"
... 
```

prateek: suggestion(take/leave): iirc this init will be invoked for all test runs (not just benchmark executions). might be better to do this as the first step of the Benchmark... func to avoid the extra time for all unit tests, could use something like https://blog.golang.org/subtests#TOC_6. to make it more manageable

jeromefroe: The Go toolchain guarantees that [test binaries will be executed with the working directory set to the source directory of the package the test is in](https://dave.cheney.net/2016/05/10/test-fixtures-in-go) so I believe the convention is to actually [use relative paths from that directory](https://github.com/golang/go/blob/964639cc338db650ccadeafb7424bc8ebb2c0f6c/src/compress/flate/reader_test.go#L25).

jeromefroe: Good point, I'll set them as subtests so we can perform the initialization only when these benchmarks are run.

prateek: My main concern was if you moved this snippet of code around (up/down) a package, it'd be more robust to copy/paste if use the absolute path. Either way is fine, up to you.

PR #17, opened 2018-02-16
Refactor package structure and interfaces

jeromefroe:
This diff refactors the package structure of m3ninx and some of the interfaces. Will write up a better description summarizing the changes once this is no longer a WIP.

cc @prateek


idx/index/segment/types.go
```diff
@@ -0,0 +1,37 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package segment
+
+import (
+	"github.com/m3db/m3ninx/idx/index"
+)
+
+// Segment ...
```
prateek: nit: fill comment out

idx/index/types.go
```diff
@@ -0,0 +1,129 @@
...
+
+package index
+
+import (
+	"github.com/m3db/m3ninx/idx/doc"
+)
+
+// Index is a collection of searchable documents.
+type Index interface {
+	Writer
```
prateek: does an Index have to be writable? e.g. when TTL'ing in m3db, we'll have multiple immutable indexes, and one mutable index, with whatever types of segments internal to each.

idx/doc/types.go
```diff
@@ -54,11 +48,18 @@ type Document struct {
 	Fields []Field
 }
 
-// HashSize is the number of the hashed ID.
-const HashSize = 16
+// Iterator provides an iterator over a collection of documents.
+type Iterator interface {
```
prateek: +1

idx/index/types.go
```diff
@@ -0,0 +1,129 @@
...
+	// specified fields are returned in the documents.
+	Docs(pl PostingsList, fields [][]byte) (doc.Iterator, error)
+}
+
+// DocID is the unique identifier of a document in the index.
+// TODO: Use a uint64, currently we only use uint32 because we use
+// Roaring Bitmaps for the implementation of the postings list.
+type DocID uint32
+
+// PostingsList is a collection of docIDs. The interface only supports
```
prateek: should the various PostingsList types be moved to the `postings` package now?

idx/index/segment/mem/types.go
```diff
@@ -18,25 +18,31 @@
...
+// Options is a collection of knobs for an in-memory segment.
 type Options interface {
 	// SetInstrumentOptions sets the instrument options.
 	SetInstrumentOptions(value instrument.Options) Options
 
 	// InstrumentOptions returns the instrument options.
 	InstrumentOptions() instrument.Options
+
+	// SetPostingsListPool sets the PostingsListPool.
+	SetPostingsListPool(value segment.PostingsListPool) Options
```
prateek: should this be `postings.Pool` now?

idx/index/segment/mem/types.go
```diff
@@ -18,25 +18,31 @@
...
+
+	// PostingsListPool returns the PostingsListPool.
+	PostingsListPool() segment.PostingsListPool
```
prateek: here too?

idx/index/types.go
```diff
@@ -0,0 +1,129 @@
...
+	// guaranteed to be searchable once the Insert method returns.
+	Insert(d doc.Document) error
+
+	// Closes closes the index.
+	Close() error
+}
+
+// Snapshot is a point-in-time snapshot of an index. It must be closed to
+// release the internal readers.
+type Snapshot interface {
```
prateek: I think you'll need a couple more methods here:
- Ability to go from DocID(s)/PostingsList -> Documents
- Ability to iterate known documents in the snapshot

idx/search/types.go
```diff
@@ -0,0 +1,40 @@
...
+
+package search
+
+import (
+	"github.com/m3db/m3ninx/idx/doc"
+	"github.com/m3db/m3ninx/idx/index"
+)
+
+// Query is a search query for documents.
+type Query interface {
```
prateek: Thought about our chat on this some more. Feels like it'd be a better abstraction to have a Query composed of some other type (e.g. Filter/Predicate). And that type has the `Execute` method. That way the searcher is responsible for the composition of results, where as the sub-type can be delegated the responsibility of executing a type of query. 

What do you think?

index/segment/mem/results_iter.go
```diff
@@ -1,134 +0,0 @@
-// Copyright (c) 2017 Uber Technologies, Inc.
```
prateek: might be useful to migrate to the implementation of `doc.Iterator`?

idx/index/types.go
jeromefroe: I modified this interface slightly to reflect that. It now includes a `Seal` method which will mark the index as immutable. Currently I have the `Seal` method returning an iterator but I could also envision it returning a `Snapshot` reflecting the state of the index after it was closed. Personally I'm leaning towards the latter so that we're explicit about the Index no longer being writable as opposed to expecting the client to maintain state on which index is sealed or not themselves.

idx/index/types.go
jeromefroe: Hmm, I wonder if that wouldn't be better left entirely to the Readers? For example, the first method is provided by the `Docs` method on a `Reader`. There's no method for the second though it could be added. I'm curious why you feel there's a need to iterate over all the documents known to the `Reader`? Said another way, I'm just not aware of a use case that needs the ability to iterate over all documents instead of being to just search over them with a query.

idx/search/types.go
jeromefroe: I'm not sure I completely understand what you mean here. Queries can be composed of other queries, for example the `conjuctionQuery` is composed of a slice of queries and is responsible for executing them and taking their intersection so I'm not sure I see what the benefit of another type would be.

prateek: The difference in the two approaches being without `Filter`(or w/e), `Query.Execute` has to execute all its sub-components (or atleast that's what the signature would imply). The whole point of the `Searcher` interface is to do that, no?

idx/index/types.go
prateek: Per our offline chat, either Index/Segment will need the Iterator to allow for other sub-systems to interact with them. e.g.  The type responsible for persisting a segment/index to disk would use this.

idx/index/types.go
prateek: I don't think Seal() can return an Iterator because you still need to be able to query the resulting type. 

Returning a Snapshot would work as long as we could query it (think you already have the types for this), and iterate the documents in it.

prateek: Could we just return a segment instead?

prateek:
Could you sketch out how you envision the new types (Index/Snapshot/Searcher/Query) interact with Segments?

jeromefroe:
Hey, sorry for the delayed response, here's what I have in mind (can add this to the repo as well once things solidify):

```
┌─────────────────────────────────────────────────────┐                     
│                        Index                        │                     
│                                                     │                     
│ ┌──────────┐ ┌──────────┐ ┌──────────┐ ┌──────────┐ │                     
│ │          │ │          │ │          │ │          │ │                     
│ │          │ │          │ │          │ │          │ │                     
│ │          │ │          │ │          │ │          │ │                     
│ │          │ │          │ │          │ │          │ │                     
│ │          │ │          │ │          │ │          │ │                     
│ │ Mutable  │ │Immutable │ │Immutable │ │Immutable │ │                     
│ │ Segment  │ │ Segment  │ │ Segment  │ │ Segment  │ │                     
│ │          │ │          │ │          │ │          │ │                     
│ │          │ │          │ │          │ │          │ │                     
│ │          │ │          │ │          │ │          │ │                     
│ │          │ │          │ │          │ │          │ │                     
│ │          │ │          │ │          │ │          │ │                     
│ └──────────┘ └──────────┘ └──────────┘ └──────────┘ │                     
└───────┬────────────┬────────────┬────────────┬──────┘                     
        │            │     ║      │            │                            
        │            │     ▼      │            │                            
┌───────┼────────────┼────────────┼────────────┼──────┐                     
│       ▼            ▼ Snapshot   ▼            ▼      │   ┌────────────────┐
│ ┌──────────┐ ┌──────────┐ ┌──────────┐ ┌──────────┐ │   │                │
│ │  Reader  │ │  Reader  │ │  Reader  │ │  Reader  │ │   │     Query      │
│ └──────────┘ └──────────┘ └──────────┘ └──────────┘ │   │                │
└─────────────────────────────────────────────────────┘   └────────────────┘
                           │                                       │        
                           │                                       │        
                           │                                       │        
                           │         ┌─────────────────────┐       │        
                           │         │                     │       │        
                           └────────▶│      Searcher       │◀──────┘        
                                     │                     │                
                                     └─────────────────────┘                
                                                │                           
                                                ▼                           
                                     ┌─────────────────────┐                
                                     │      Documents      │                
                                     └─────────────────────┘                
```

In words,

- An Index is composed of multiple segments, only one of which is actively being written to.
- Each segment can provide a Reader which allows consistent access to the documents in a segment at the point in time that the Reader is created.
- A Snapshot is a collection of Readers. I still need to flesh this out more since there's some questions I'm unsure of (should the caller be responsible for taking a new Snapshot periodically or should the Snapshot be responsible for the lifecycle of the Readers, closing expired ones and getting new Readers after writes come in, if its the latter we probably want to change the name).
- A Query encapsulates a query to the index (e.g. `service:foo name:requests`). It knows how to execute itself against a single Reader.
- The Searcher is used to execute a Query against each Reader in a Snapshot and return the results.

postings/roaring.go
```diff
@@ -68,21 +69,21 @@ func (d *roaringPostingsList) Intersect(other ImmutablePostingsList) error {
...
-func (d *roaringPostingsList) Difference(other ImmutablePostingsList) error {
+func (d *roaringPostingsList) Difference(other List) error {
 	o, ok := other.(*roaringPostingsList)
 	if !ok {
 		return errDifferenceRoaringOnly
 	}
 
-	d.Lock()
 	o.RLock()
+	d.Lock()
```
prateek: curious why you changed the order of the locks?

index/segment/mem/iterator.go
```diff
@@ -0,0 +1,95 @@
...
+	segment      readableSegment
+	postingsIter postings.Iterator
+	wg           *sync.WaitGroup
+
+	current doc.Document
+	err     error
+	closed  bool
+}
+
+func newIterator(s readableSegment, pi postings.Iterator, wg *sync.WaitGroup) doc.Iterator {
```
prateek: what's the `wg` for?

index/segment/mem/reader.go
```diff
@@ -0,0 +1,128 @@
...
+
+	segment readableSegment
+	maxID   postings.ID
+	wg      *sync.WaitGroup
+
+	closed bool
+}
+
+func newReader(s readableSegment, maxID postings.ID, wg *sync.WaitGroup) index.Reader {
+	wg.Add(1)
```
prateek: same here, what's the wg for?

index/segment/mem/segment.go
```diff
@@ -0,0 +1,258 @@
...
+package mem
+
+import (
+	"errors"
+	"regexp"
+	"sync"
+
+	"github.com/m3db/m3ninx/doc"
+	"github.com/m3db/m3ninx/index"
+	sgmt "github.com/m3db/m3ninx/index/segment"
```
prateek: any reason to alias here?

index/segment/mem/segment.go
```diff
@@ -0,0 +1,258 @@
...
+	docs struct {
+		sync.RWMutex
+
+		data []doc.Document
+	}
+
+	// Current writer and reader IDs. Writers increment the writer ID for each new
+	// document and only increment the reader ID after the document has been fully
+	// indexed by the segment. Readers do not need to acquire a lock.
+	ids struct {
```
prateek: why have a lock at all if they're atomics?

index/segment/mem/segment.go
```diff
@@ -0,0 +1,258 @@
...
+
+func (s *segment) Reader() (index.Reader, error) {
+	s.state.RLock()
+	if s.state.closed {
+		s.state.RUnlock()
+		return nil, errSegmentClosed
+	}
+
+	maxID := s.ids.reader.Load()
+	r := newReader(s, maxID, &s.wg)
```
prateek: ah you're using the wg for ref counting? how about using https://github.com/m3db/m3x/blob/master/checked/types.go#L81 instead?

search/types.go
```diff
@@ -0,0 +1,41 @@
...
+package search
+
+import (
+	"github.com/m3db/m3ninx/doc"
+	"github.com/m3db/m3ninx/index"
+	"github.com/m3db/m3ninx/postings"
+)
+
+// Query is a search query for documents.
+type Query interface {
```
prateek: I don't understand the distinction between a query and a searcher. Will follow up offline

index/index_mock.go
```diff
@@ -0,0 +1,225 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Code generated by MockGen. DO NOT EDIT.
+// Source: github.com/m3db/m3ninx/index/types.go
+
+package index
+
+import (
+	gomock "github.com/golang/mock/gomock"
```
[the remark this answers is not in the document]
jeromefroe: Sure thing. I actually ended up doing that for the semaphore diff, so I'll just rebase onto master to pull them in.

postings/roaring.go
jeromefroe: This was just to be consistent with the other methods which acquire `o` first and then `d`.

index/segment/mem/iterator.go
jeromefroe: As you point out below, it's to do ref counting.

index/segment/mem/segment.go
jeromefroe: This was just so I could reuse `segment` for the name of the struct.

index/segment/mem/segment.go
jeromefroe: If we had only a single writer then the lock would be unnecessary since only that writer would be responsible for incrementing the IDs. With multiple writers the lock is used to enforce that only a single writer can increment both. I'll probably come back to this at some point. The main difficulty that needs to be solved is that with multiple writers we don't want to set the ID to a given value if documents with lower IDs haven't finished being indexed yet.

index/segment/mem/segment.go
jeromefroe: We can use `Ref`, but it seems a bit heavyweight at the moment so I think it might be simpler to stick with just the wait group for now.

index/segment/mem/postings_map.go
```diff
@@ -0,0 +1,113 @@
...
+		p = m.opts.PostingsListPool().Get()
+	}
+	return p
+}
+
+func (m *postingsMap) getRegex(re *regexp.Regexp) []postings.List {
+	m.RLock()
+
+	initLen := math.Min(regexpMatchFactor*float64(len(m.postings)), regexMatchMaxLen)
+	ps := make([]postings.List, 0, int(initLen))
```
prateek: Why get back a list of postingsLists? Possible to combine into one?

index/types.go
```diff
@@ -21,22 +21,96 @@
...
+	// regular expression.
+	MatchRegexp(field, regexp []byte, compiled *regexp.Regexp) (postings.List, error)
+
+	// Docs returns an iterator over the documents whose IDs are in the provided
+	//  postings list.
+	Docs(pl postings.List) (doc.Iterator, error)
+
+	// Clone returns an independent clone of this Reader. The returned reader must be
+	// closed independently of this Reader.
+	Clone() (Reader, error)
```
prateek: How is the `Clone()` method used?

index/segment/mem/segment.go
prateek: OK, then how about not using atomic.ID for them? It's confusing to see an atomic guarded by a lock.

index/types.go
```diff
@@ -21,22 +21,96 @@
...
+	}
+	return multiErr
 }
 
-// Options is a set of knobs by which to tweak Index-ing behaviour.
-type Options interface {
-	// SetInstrumentOptions sets the instrument options.
-	SetInstrumentOptions(value instrument.Options) Options
+// Clone returns a clone of all of the internal Readers.
+func (rs Readers) Clone() (Readers, error) {
```
prateek: Where is this used?

search/types.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/m3db/m3ninx/index"
+	"github.com/m3db/m3ninx/postings"
+
+	"github.com/m3db/m3x/errors"
+)
+
+// Query is a search query for documents.
+type Query interface {
+	// Searcher returns a Searcher for executing the query over a collection of Readers.
+	Searcher(rs index.Readers) (Searcher, error)
```
prateek: talked offline: change to accept snapshot

index/segment/mem/reader.go
```diff
@@ -0,0 +1,117 @@
...
+type reader struct {
+	sync.RWMutex
+
+	segment ReadableSegment
+	maxID   postings.ID
+
+	closed bool
+}
+
+func newReader(s ReadableSegment, maxID postings.ID) index.Reader {
```
prateek: add methods for pooling in a subsequent PR: e.g. Reset(Segment, maxID) ...

index/segment/mem/postings_map.go
prateek: i can see the flexibility of returning multiple postings lists too, maybe it makes sense to return as an iterator of postignslists?

index/segment/mem/segment.go
jeromefroe: Talked offline, will flesh out the comment to explain the semantics more.

index/segment/mem/postings_map.go
jeromefroe: This method is only used in one place at the moment, if we start to use it more often I think we can make that change, but at the moment I think it would be simpler to just return the slice.

index/types.go
jeromefroe: Talked offline, since a Reader maintains a reference count to the segment it was created from, `Clone` is used to increment the reference count and return a copy of the given Reader.

index/types.go
jeromefroe: Talked offline, this is used in the `search` package when we want to have multiple independent versions of the same Reader.

search/types.go
jeromefroe: Thought about this a little more, I'm not sure we want to pass the `Snapshot` here because then the lifecycle of the `Snapshot` becomes tied to the `Searcher`s. I think it would be preferable to have the layer above manage the `Snapshot` exclusively. That way it would know when the `Snapshot` is open or closed and be able to get `Reader`s from it as appropriate.

jeromefroe: Actually, changed my mind, I like passing the Snapshot because we only need a reference to it and not ownership, will update.

search/searcher/term.go
```diff
@@ -0,0 +1,86 @@
...
+	idx  int
+	curr postings.List
+
+	closed bool
+	err    error
+}
+
+// NewTermSearcher returns a new searcher for finding documents which match the given term.
+// It is not safe for concurrent access.
+func NewTermSearcher(rs index.Readers, field, term []byte) search.Searcher {
```
prateek: nit: should you take a `Snapshot` as input here? that way you don't need to clone the readers slice, can reference count on the Snapshot itself

search/searcher/regexp.go
```diff
@@ -0,0 +1,90 @@
...
+	idx  int
+	curr postings.List
+
+	closed bool
+	err    error
+}
+
+// NewRegexpSearcher returns a new searcher for finding documents which match the given regular
+// expression. It is not safe for concurrent access.
+func NewRegexpSearcher(rs index.Readers, field, regexp []byte, compiled *re.Regexp) search.Searcher {
```
prateek: same here

search/searcher/empty.go
```diff
@@ -0,0 +1,75 @@
...
+
+	idx int
+
+	closed bool
+	err    error
+}
+
+// NewEmptySearcher returns a new searcher which always returns an empty postings list.
+// It is not safe for concurrent access.
+func NewEmptySearcher(len int) search.Searcher {
```
prateek: this is just for testing?

search/searcher/disjunction.go
```diff
@@ -0,0 +1,108 @@
...
+	idx  int
+	curr postings.List
+
+	closed bool
+	err    error
+}
+
+// NewDisjunctionSearcher returns a new Searcher which matches documents which are matched
+// by any of the given Searchers. It is not safe for concurrent access.
+func NewDisjunctionSearcher(ss search.Searchers) (search.Searcher, error) {
```
prateek: not sure I understand the intent of this code. 

Here's my take: for input it takes a series of searchers, which are effectively iterators over postings lists. Then you take the union of these indexwise. 

i.e. 
```
// Given input: 
  S1: [P_11, P_12, ...]
  S2: [P_21, P_22, ...]

// The output this gives back is:
  [P_11 ∪ P_21, P_12 ∪ P_22, ... ]
```

Is that right?

search/searcher/conjunction.go
```diff
@@ -0,0 +1,113 @@
...
+	idx  int
+	curr postings.List
+
+	closed bool
+	err    error
+}
+
+// NewConjunctionSearcher returns a new Searcher which matches documents which match each
+// of the given Searchers. It is not safe for concurrent access.
+func NewConjunctionSearcher(ss search.Searchers) (search.Searcher, error) {
```
prateek: same question as for the disjunction. Why do it piecemeal instead of all at once

search/searcher/conjunction.go
```diff
@@ -0,0 +1,113 @@
...
+	var length int
+	if len(ss) > 0 {
```
prateek: looks like you could share this code across the conjunction/disjunction ctors. maybe add:

```
type validatedSearchers search.Searchers

func (ss validatedSearchers) Validate() error {
  var length int 
  ... 
}
```

search/query/boolean.go
```diff
@@ -0,0 +1,101 @@
...
+// booleanQuery is a compound query composed of must, should, and must not queries.
+type booleanQuery struct {
+	must    search.Query
+	should  search.Query
+	mustNot search.Query
+}
+
+// NewBooleanQuery returns a new compound query such that:
+//   - resulting documents match ALL of the must queries
+//   - resulting documents match AT LEAST ONE of the should queries
```
prateek: i don't follow the rationale for `should`, do you have a use-case in mind for it?

search/types.go
prateek: Sounds good.

index/segment/mem/postings_map.go
prateek: Sure

prateek:
@jeromefroe looking good. Left last few questions, should be good to land after this round.

search/searcher/term.go
jeromefroe: I think it would be simpler if we only reference counted the segments. That's why we would take a reference count on the Snapshot, we would transitively being taking a reference count on the segment.

search/searcher/empty.go
jeromefroe: No, this is the searcher returned when no queries are passed to a conjunction or disjunction query.

search/searcher/disjunction.go
jeromefroe: That's correct, it's meant to be segment-wise though. Doing it across indices would require a little more care since we lose postings list uniqueness.

search/query/boolean.go
jeromefroe: I'm gonna get rid of this. Boolean queries are common in general IR where documents can be ranked but make less sense for us where whether a document matches a query is a binary yes or no.

search/types.go
```diff
@@ -0,0 +1,74 @@
...
+func (ss Searchers) Close() error {
+	multiErr := errors.NewMultiError()
+	for _, s := range ss {
+		err := s.Close()
+		if err != nil {
+			multiErr = multiErr.Add(err)
+		}
+	}
+
+	if multiErr.Empty() {
```
prateek: can simplify the if + fallback to 

```
  return multiError.FinalError()
```

search/searcher/empty.go
prateek: Hm why do you need `len` for that?

search/searcher/term.go
prateek: For sure its simpler to have the readers cloned but i'm concerned about the allocs & contention. 

For e.g. if you have a disjunction query with 3 sub-parts, you'd end up 4 different clone of the Readers. All the while you only needed to guarantee the segment's are valid for the duration of the top level query. Which a single clone of the readers is sufficient for.

search/types.go
```diff
@@ -0,0 +1,74 @@
...
+// Searcher executes a query against a collection of Readers. It is an iterator which
+// returns the postings lists of the documents it matches for each segment. It is not
+// safe for concurrent access.
+type Searcher interface {
+	// Next returns the whether the iterator has another postings list.
+	Next() bool
+
+	// Current returns the current postings list. It is only safe to call Current immediately
+	// after a call to Next confirms there are more postings lists remaining.
+	Current() postings.List
```
prateek: per offline chat, maybe put down a todo about conversion from postings.List -> doc.Iterator

search/searcher/disjunction.go
prateek: Fair. Could you put a note down in the interface explaining this semantic in more detail

search/searcher/term.go
jeromefroe: Talked offline, we won't be cloning the Readers anymore. The top-level `Executor` will be responsible for enforcing their lifetimes.

search/searcher/empty.go
jeromefroe: Talked offline, Searchers search over segments independently so we need to know the number of segments that this searcher is searching over.

search/searcher/disjunction.go
jeromefroe: Yep, added a note to the documentation on the interface.

search/types.go
jeromefroe: Talked offline, this will now be covered by the `Executor`.

PR #20, opened 2018-03-09
Add public idx package

jeromefroe:
This diff adds the `idx` package which will contain all the public facing interfaces that a client would interact with. I haven't updated the package structure but will likely leave until what we decide on what the `idx` package should look like.

cc @prateek

### DIFF
--- a/idx/query.go
+++ b/idx/query.go
@@ -1,0 +1,73 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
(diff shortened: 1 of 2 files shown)


idx/query.go
```diff
@@ -0,0 +1,91 @@
...
+// BooleanQueryBuilder is used to build boolean queries.
+type BooleanQueryBuilder struct {
+	must    []search.Query
+	should  []search.Query
+	mustNot []search.Query
+}
+
+// Add adds a query with the given type of clause to the builder. The builder is also
+// returned so that calls to Add can be chained together.
+func (b *BooleanQueryBuilder) Add(q Query, c BooleanClause) *BooleanQueryBuilder {
```
prateek: how about three functions instead of Add, i.e.

```
func (b *BooleanQueryBuilder) Must(q Query) *BooleanQueryBuilder {}
func (b *BooleanQueryBuilder) Should(q Query) *BooleanQueryBuilder {}
func (b *BooleanQueryBuilder) ShouldNot(q Query) *BooleanQueryBuilder {}
```

idx/query.go
```diff
@@ -0,0 +1,91 @@
...
+		query: q,
+	}, nil
+}
+
+// BooleanClause represents the type of clause in a boolean query.
+type BooleanClause int
+
+// The valid types of clauses in a boolean query are Must, Should, and MustNot.
+const (
+	Must BooleanClause = iota
```
prateek: what's the difference between `Must` and `Should`?

jeromefroe: A document has to match every `Must` query but only has to match a single `Should` query. I'm thinking about just removing this in favor of just having conjunction and disjunction queries since Boolean queries are more relevant when documents can be scored but we just need to support a binary yes or no.

idx/query.go
```diff
@@ -0,0 +1,73 @@
...
+		return Query{}, err
+	}
+	return Query{
+		query: q,
+	}, nil
+}
+
+// NewConjunctionQuery returns a new query for finding documents which match each of the
+// given queries.
+func NewConjunctionQuery(queries ...Query) (Query, error) {
```
prateek: nit: expose a version of this method which takes a slice. That way users don't have to cast down from a slice only for you to create the slice again.

idx/query.go
```diff
@@ -0,0 +1,73 @@
...
+		qs = append(qs, q.query)
+	}
+	return Query{
+		query: query.NewConjuctionQuery(qs),
+	}, nil
+}
+
+// NewDisjunctionQuery returns a new query for finding documents which match at least one
+// of the given queries.
+func NewDisjunctionQuery(queries ...Query) (Query, error) {
```
prateek: here too

idx/query.go
jeromefroe: Talked offline, we need to create a new slice because the second slice is of a different type than the first.

idx/types.go
```diff
@@ -0,0 +1,44 @@
...
+	// Searcher returns a Searcher over a point-in-time view of the index.
+	Searcher() (Searcher, error)
+
+	// Close closes the index.
+	Close() error
+}
+
+// Searcher provides search over a point-in-time view of an index.
+type Searcher interface {
+	Search(q Query) (doc.Iterator, error)
```
prateek: nit: add comments here

PR #23, opened 2018-03-15
Add implementation of executor

jeromefroe:
This diff adds a concrete implementation of the `Executor` interface.


search/executor/executor.go
```diff
@@ -19,3 +19,79 @@
...
+func (e *executor) Close() error {
+	e.Lock()
+	if e.closed {
+		e.Unlock()
+		return errExecutorClosed
+	}
+	e.closed = true
+	e.Unlock()
+
+	// Wait for all references to the executor to be released.
```
prateek: don't wait here, instead force all the references to release.

search/executor/iterator.go
```diff
@@ -0,0 +1,150 @@
...
+		refCount: rc,
+		searcher: s,
+		readers:  rs,
+	}
+
+	currIter, err := it.nextIter()
+	if err != nil {
+		return nil, err
+	}
+	rc.IncRef()
```
prateek: looks like this is the only place you incref in this file. Why have the iterator separate from the executor? their lifecycles look 1-1 to me.

search/executor/executor.go
jeromefroe: I'm not entirely sure what you mean. Require the references to be released before calling Close?

search/executor/iterator.go
jeromefroe: There's a corresponding `decRef` in the `Close` method. The executor can return multiple iterators which the client will use. Some may be used immediately and closed but others may be used in bursts, for example when processing in batches as we may do when paging so I think it would be beneficial if their lifetimes weren't tied together.

PR #25, opened 2018-04-03
Migrate to using m3x/generics for []bytes map

### LINKED ISSUE #11: Avoid using a map with string keys for the terms dictionary
String keys require extra allocs that we could avoid if we rolled with our own hashmap implementation. We should evaluate the perf impact of doing so and make the corresponding changes. 

Doing so also opens up the possibility of using a map hidden from the GC ("native"/mmap'd), which we should look into too.

prateek:
`benchcmp` between master and this branch:

```
benchcmp old.txt new.txt
benchmark                                                                  old ns/op     new ns/op     delta
BenchmarkTermsDict/benchmark_Insert_with_simple_terms_dictionary-4         1639493       1490682       -9.08%
BenchmarkTermsDict/benchmark_MatchTerm_with_simple_terms_dictionary-4      4296456       4414548       +2.75%
BenchmarkTermsDict/benchmark_MatchRegex_with_simple_terms_dictionary-4     92783         78484         -15.41%

benchmark                                                                  old allocs     new allocs     delta
BenchmarkTermsDict/benchmark_Insert_with_simple_terms_dictionary-4         14899          4919           -66.98%
BenchmarkTermsDict/benchmark_MatchTerm_with_simple_terms_dictionary-4      40008          40008          +0.00%
BenchmarkTermsDict/benchmark_MatchRegex_with_simple_terms_dictionary-4     95             95             +0.00%

benchmark                                                                  old bytes     new bytes     delta
BenchmarkTermsDict/benchmark_Insert_with_simple_terms_dictionary-4         390756        317038        -18.87%
BenchmarkTermsDict/benchmark_MatchTerm_with_simple_terms_dictionary-4      5931839       5931696       -0.00%
BenchmarkTermsDict/benchmark_MatchRegex_with_simple_terms_dictionary-4     3734          3734          +0.00%
```

fixes #11


index/segment/mem/fieldsmap/map.go
```diff
@@ -0,0 +1,254 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package fieldsmap
```
jeromefroe: nit: How do you feel about simplifying this to `fields` so we can have `fields.Map`?

index/segment/mem/postingsmap/concurrent_map.go
```diff
@@ -0,0 +1,148 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package postingsmap
```
jeromefroe: nit: How about `postings` here so we can just have `postings.ConcurrentMap`?

index/segment/mem/postingsmap/concurrent_map.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// newInternalMap returns a new []bytes->postings.MutableList map.
+func newInternalMap(initialSize int) *Map {
+	return newMap(mapOptions{
+		hash: func(k []byte) MapHash {
+			return MapHash(xxhash.Sum64(k))
+		},
+		equals:      bytes.Equal,
+		copy:        undefinedCopyFn,
```
jeromefroe: nit: same here, any reason to not just pass nil or leave undefined?

index/segment/mem/fieldsmap/new_map.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+// New returns a new []bytes->*postingsmap.ConcurrentMap.
+func New(initialSize int) *Map {
+	return newMap(mapOptions{
+		hash: func(k []byte) MapHash {
+			return MapHash(xxhash.Sum64(k))
+		},
+		equals:      bytes.Equal,
+		copy:        undefinedCopyFn,
```
jeromefroe: nit: any reason to not just pass nil or leave undefined?

index/segment/mem/terms_dict_bench_test.go
```diff
@@ -82,29 +70,11 @@ func BenchmarkTermsDict(b *testing.B) {
 }
 
 func benchmarkInsertSimpleTermsDict(docs []doc.Document, b *testing.B) {
+	dict := newSimpleTermsDict(NewOptions().SetInitialCapacity(10))
 	b.ReportAllocs()
+	b.ResetTimer()
 
 	for n := 0; n < b.N; n++ {
-		b.StopTimer()
-		dict := newSimpleTermsDict(NewOptions())
```
jeromefroe: If we move the initialization out of the benchmark won't every iteration after the first be a no-op?

index/segment/mem/terms_dict_bench_test.go
```diff
@@ -114,79 +84,35 @@ func benchmarkInsertTrigramTermsDict(docs []doc.Document, b *testing.B) {
 }
 
 func benchmarkMatchTermSimpleTermsDict(docs []doc.Document, b *testing.B) {
-	b.ReportAllocs()
```
jeromefroe: nit: `ReportAllocs` only affects the output of the test to include allocation information automatically. Perhaps its worth leaving it as a setup step just to be consistent?

Makefile
```diff
@@ -83,8 +87,21 @@ proto-gen: install-proto-bin install-license-bin
 	@echo Generating protobuf files
 	PACKAGE=$(m3ninx_package) $(auto_gen) $(proto_output_dir) $(proto_rules_dir)
 
+.PHONY: install-genny
```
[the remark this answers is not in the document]
xichen2020: Btw I already added this target as `install-generics-bin` to `ci-script`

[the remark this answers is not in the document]
prateek: cheers, will switch to it.

index/segment/mem/fieldsmap/map.go
prateek: sure

index/segment/mem/postingsmap/concurrent_map.go
prateek: sure

index/segment/mem/postingsmap/concurrent_map.go
prateek: Both exhibit the same behaviour but I prefer the explicitness of this. When someone traces the panic to the code in this version, they find the comment declaring the panic is intentional. In the other version, the panic would be in generated code, which is a little too implicit IMO.

index/segment/mem/fieldsmap/new_map.go
prateek: same as above

index/segment/mem/fieldsmap/map.go
prateek: renamed to `fieldsgen`

index/segment/mem/postingsmap/concurrent_map.go
prateek: renamed to `postingsgen`

index/segment/mem/terms_dict_bench_test.go
prateek: Fair. I thought you were testing something else (allocs for repeated inserts). Will undo change.

index/segment/mem/terms_dict_bench_test.go
prateek: fair enough

Makefile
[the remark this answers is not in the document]
prateek: done

generated/generics/generate.sh
```diff
@@ -0,0 +1,26 @@
...
+
+GENERIC_MAP_PATH=${GOPATH}/src/github.com/m3db/m3ninx/vendor/github.com/m3db/m3x/generics/hashmap
+GENERIC_MAP_IMPL=${GENERIC_MAP_PATH}/map.go
+
+if [ ! -f "$GENERIC_MAP_IMPL" ]; then
+  echo "${GENERIC_MAP_IMPL} does not exist"
+  exit 1
+fi
+
+GENERATED_PATH=${GOPATH}/src/github.com/m3db/m3ninx/index/segment/mem
```
jeromefroe: It seems this is pretty specific to the current packages we're generating, is it worth having perhaps just a short README to explain how to add another package in the future?

index/segment/mem/fieldsgen/new_map.go
```diff
@@ -0,0 +1,52 @@
...
+			return MapHash(xxhash.Sum64(k))
+		},
+		equals:      bytes.Equal,
+		copy:        undefinedCopyFn,
+		finalize:    undefinedFinalizeFn,
+		initialSize: initialSize,
+	})
+}
+
+var undefinedCopyFn CopyFn = func([]byte) []byte {
```
jeromefroe: nit: any reason to just not include this in the options so it gets the default value of nil?

prateek: IMO the cause of a crash will be more apparent if we raise a panic pointing to a piece of code with a comment saying it's intentional that it's not implemented. If we go the other route (leaving nil), the panic would happen in generated code, which would make figuring the same thing out a little harder.

generated/generics/generate.sh
prateek: Sure, I put down a few paragraphs in the script file.

PR #26, opened 2018-04-03
Include metalinter config

prateek:


.metalinter.json
```diff
@@ -0,0 +1,27 @@
...
+    "varcheck",
+    "structcheck",
+    "goconst",
+    "ineffassign",
+    "unconvert",
+    "misspell",
+    "golint",
+    "gosimple",
+    "unused",
+    "maligned"
```
jeromefroe: Is it worth adding go vet and go fmt as well?

index/segment/mem/iterator.go
```diff
@@ -34,11 +34,12 @@ var (
...
 	postingsIter postings.Iterator
-	maxID        postings.ID
 
 	current doc.Document
-	err     error
-	closed  bool
+
+	err    error
+	closed bool
+	maxID  postings.ID
```
jeromefroe: Is this because of structcheck? If so, is it worth preferring the old layout since it had the benefit of distinguishing immutable versus mutable fields.

prateek: it's because of `maligned`: this layout is 80 bytes, the older one was 88. 

I'll put a comment to indicate why we order this way and the mutability of fields.

.metalinter.json
prateek: opened https://github.com/m3db/m3ninx/issues/27 to track

index/segment/mem/iterator.go
prateek: done

PR #28, opened 2018-04-04
[codec] Add a documents writer and reader

jeromefroe:
This diff creates the codec package for encoding and decoding the different components which comprise a segment. It also adds an implementation for a `DocumentsWriter` which is capable of encoding documents.

This is still very much in progress. TODO:

- [X] Add `DocumentsReader`
- [X] Add tests
- [X] Add README describing the file format


index/codec/types.go
```diff
@@ -0,0 +1,69 @@
...
+		return "documents"
+	default:
+		return "unknown"
+	}
+}
+
+// DocWriter is used to write a documents file.
+type DocWriter interface {
+	// Prepare prepares the DocWriter.
+	Prepare() error
```
prateek: maybe `Open` instead of `Prepare`?

index/codec/types.go
```diff
@@ -0,0 +1,69 @@
...
+	switch t {
+	case DocumentsFile:
+		return "documents"
+	default:
+		return "unknown"
+	}
+}
+
+// DocWriter is used to write a documents file.
+type DocWriter interface {
```
prateek: do you want to create a `DocReader` too?

index/codec/doc.go
```diff
@@ -0,0 +1,304 @@
...
+		return err
+	}
+	if c != r.checksum.get() {
+		return errChecksumMismatch
+	}
+	r.checksum.reset()
+	return nil
+}
+
+func (r *docReader) Read() (doc.Document, error) {
```
prateek: maybe using this would look better as an iterator?

index/codec/types.go
```diff
@@ -0,0 +1,69 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package codec
```
prateek: `codec` implies compression, would `encoding` or `serialize` be more appropriate?

index/codec/doc.go
prateek: talked offline, more consistent with code elsewhere (m3db/et all) to use current impl

index/codec/types.go
prateek: talked offline - `serialize`

prateek:
talked offline: 
- add public ctors to these types
- expose a segment -> doc.Iterator method

index/encoding/doc/doc.go
```diff
@@ -0,0 +1,274 @@
...
+	w.enc.PutUint64(w.docs)
+	if err := w.write(); err != nil {
+		return err
+	}
+
+	w.enc.PutUint32(w.checksum.Get())
+	if err := w.write(); err != nil {
+		return err
+	}
+
```
prateek: Should this also close the underlying `w.writer`? that way the user can create an `io.Writer` and passing it to the `encoding/doc/Writer` transfers ownership of it

index/encoding/doc/doc.go
```diff
@@ -0,0 +1,274 @@
...
+	data []byte
+	dec  *encoding.Decoder
+
+	curr     uint64
+	total    uint64
+	checksum *encoding.Checksum
+}
+
+// NewReader returns a new Reader for deserializing documents from a byte slice.
+func NewReader(data []byte) Reader {
```
prateek: This looks like you need to read the entire dataset into memory before iterating on it. Anyway to stream the data as you read it?

jeromefroe:
I'll include the README in this one. Working on it at the moment.

index/encoding/doc/doc.go
```diff
@@ -0,0 +1,294 @@
...
+const (
+	magicNumber = 0x6D33D0C5
+
+	fileFormatVersion = 1
+
+	headerSize = 0 +
+		4 + // magic number
+		4 + // version
+		4 + // checksum of header fields
+		0
```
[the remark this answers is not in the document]
jeromefroe: Yea, this is just for aesthetics.

index/encoding/doc/doc.go
jeromefroe: At the moment, I'm of the feeling that we should handle the lifecycle of the `io.Writer` outside of the `doc.Writer` in case we want to, for example, encode multiple writers (documents, terms dictionary, postings list) into the same file. I don't feel strongly here though, and I think actually using the writer in practice will help decide here.

prateek: I don't think that'll happen - more inclined to making a filetype for each type of writer (separate ones for documents v terms dictionary v postings lists).

prateek: I think it'll be simpler to use if the doc.Writer takes ownership of the io.Writer. That way we only need to track one var's lifecycle instead of two.

PR #32, opened 2018-04-17
[index] Support document IDs

jeromefroe:
This PR adds native support for IDs in a segment.


doc/document.go
```diff
@@ -0,0 +1,115 @@
...
+
+// Equal returns whether this document is equal to another.
+func (d Document) Equal(other Document) bool {
+	if len(d.Fields) != len(other.Fields) {
+		return false
+	}
+
+	l, r := Fields(d.Fields), Fields(other.Fields)
+
+	// Make a shallow copy of the Fields so we don't mutate the document.
```
prateek: would it be better to do the `O(n^2)` version? i think it'd require fewer overall comparisons

jeromefroe: This function is only used for testing currently and I don't see it ever being in the hot path so I think the simplest approach might be the best for now. I personally thought this was clear but don't have strong feelings.

doc/document.go
```diff
@@ -0,0 +1,115 @@
...
+
+package doc
+
+import (
+	"bytes"
+	"sort"
+)
+
+// IDFieldName is the field name reserved for IDs.
+var IDFieldName = []byte("_id")
```
prateek: nit: rename to indicate it's a reserved word. `ReservedFieldNameID` or something along the lines.

Assuming it's a var to allow library users to change, yea? if so, could you put a little note indicating that.

index/segment/mem/segment.go
```diff
@@ -93,26 +97,60 @@ func NewSegment(offset postings.ID, opts Options) (sgmt.MutableSegment, error) {
...
+			docID = f.Value
+		}
+
 		if !utf8.Valid(f.Value) {
 			return fmt.Errorf("document contains invalid field value: %v", f.Value)
 		}
 	}
 
+	if docID != nil {
+		pl, err := s.matchTerm(doc.IDFieldName, docID)
```
prateek: couple of nits about this: 
- don't need to grab the lock in matchTerm for this codepath
- don't need to allocate a postings list for this case either (termsDict.matchTerm) does so

Could you create a separate code path which is lighter weight (`segment.containsTermWithLock` and `termsDict.containsField`)?

doc/document.go
jeromefroe: sure, sounds good

and nah, it has to be a `var` if i want it to be a byte slice. I could make it a string and make it const but then I have to convert it to a byte slice whenever I want to use it.

index/segment/mem/segment.go
```diff
@@ -93,26 +97,60 @@ func NewSegment(offset postings.ID, opts Options) (sgmt.MutableSegment, error) {
...
 	}
 
 	if s.state.sealed {
-		s.state.RUnlock()
 		return errSegmentSealed
 	}
 
-	// Validate that the document contains only valid UTF-8.
+	// Check if the document contains a document ID and ensure it's fields are valid UTF-8.
+	var docID []byte
```
prateek: could you decouple the validation into a separate function that can be used upstream too?

Something like:
```
func Validate(d doc.Document) error {
  // do all `pure` Field validation (w/o touching any of the segment internals
}

func (s *segment) InsertValidated(d doc.Document) error {
  // skip any validation, only do insertion 
}

func (s *segment) Insert(d doc.Document) error {
 // do validation
 // and then insert
}

File: index/segment/mem/segment.go [hunk 5]
```diff
@@ -93,26 +97,60 @@ func NewSegment(offset postings.ID, opts Options) (sgmt.MutableSegment, error) {
...
+		pl, err := s.matchTerm(doc.IDFieldName, docID)
+		if err != nil {
+			return fmt.Errorf("could not determine if index already contains document ID '%s'", docID)
+		}
+		if !pl.IsEmpty() {
+			// Index already contains this document.
+			return nil
+		}
+	} else {
+		uuid, err := util.NewUUID()
```
prateek: I can see the value behind this for certain use-cases. Mind hiding this behind an Option?

index/segment/mem/segment.go
jeromefroe: sounds good

doc/document.go
prateek: ah ok, if its only used in tests, mind moving this logic to a matcher to make it explicit. Don't want people accidentally using these methods if they're not intended for public use.

doc/document.go
```diff
@@ -0,0 +1,167 @@
...
+)
+
+var (
+	errMultipleIDs   = errors.New("document cannot contain multiple IDs")
+	errZeroLengthID  = errors.New("document ID cannot be of length zero")
+	errEmptyDocument = errors.New("document cannot be empty")
+)
+
+// IDReservedFieldName is the field name reserved for IDs.
+var IDReservedFieldName = []byte("_id")
```
prateek: wanted to you intended this is OK to be changed by a downstream library user. that's good yea?

index/segment/mem/segment.go
```diff
@@ -93,24 +95,38 @@ func NewSegment(offset postings.ID, opts Options) (sgmt.MutableSegment, error) {
...
+			return fmt.Errorf("could not lookup document ID '%s'", docID)
+		}
+		if ok {
+			// Index already contains this document.
+			return nil
 		}
-		if !utf8.Valid(f.Value) {
-			return fmt.Errorf("document contains invalid field value: %v", f.Value)
+	} else {
+		uuid, err := s.newUUIDFn()
```
prateek: sorry should have been clearer, I meant optionally allow users to configure whether a uuid should be autogen'd or if this should be an error.

index/segment/mem/segment.go
```diff
@@ -93,24 +95,38 @@ func NewSegment(offset postings.ID, opts Options) (sgmt.MutableSegment, error) {
...
 	if s.state.sealed {
-		s.state.RUnlock()
 		return errSegmentSealed
 	}
 
-	// Validate that the document contains only valid UTF-8.
-	for _, f := range d.Fields {
-		if !utf8.Valid(f.Name) {
-			return fmt.Errorf("document contains invalid field name: %v", f.Name)
+	docID, err := d.Validate()
```
prateek: could you also expose the path that skips validation prior to insert?

doc/document.go
```diff
@@ -0,0 +1,167 @@
...
+		if !bytes.Equal(l[i].Value, r[i].Value) {
+			return false
+		}
+	}
+
+	return true
+}
+
+// Validate validates the given document and returns its ID if it has one.
+func (d Document) Validate() ([]byte, error) {
```
prateek: feels hacky to couple validate and id retrieval into the same method. could you expose separate functions for them (possibly reusing an unexported method internally)

index/segment/mem/terms_dict.go
```diff
@@ -53,19 +53,37 @@ func (d *termsDict) Insert(field doc.Field, id postings.ID) error {
 	return postingsMap.Add(field.Value, id)
 }
 
+func (d *termsDict) ContainsTerm(field, term []byte) (bool, error) {
+	_, found, err := d.matchTerm(field, term)
```
prateek: This still allocates if not found. could you re-work to avoid that.

index/segment/mem/segment.go
jeromefroe: oh, hmm, I can't really think of when that would be the case and if it is necessary I think it's fair to say it should be handled at a layer above the segment for now

doc/document.go
jeromefroe: No, I don't think this should be ever be changed, but it can be if the caller so chooses. It's buyer-beware though, we don't offer any guarantees if they do.

doc/document.go
jeromefroe: Yea, so I did this to save an extra iteration through the fields, can move the id selection into a separate function, don't feel strongly here

index/segment/mem/segment.go
jeromefroe: As in have an option to not validate?

index/segment/mem/terms_dict.go
jeromefroe: I'm not sure I see how it still allocates if not found?

doc/document.go
prateek: talked offline, changing value to `var ... = "_m3ninx_id" `

index/segment/mem/terms_dict.go
prateek: you're right. `matchTerm != MatchTerm` :|

doc/document.go
jeromefroe: `_m3ninx_id`

index/segment/mem/segment.go
jeromefroe: make an error function

index/segment/mem/segment.go
jeromefroe: leave for now

doc/document.go
jeromefroe: leave for now

PR #34, opened 2018-04-20
[index] Add Batch method for inserting multiple documents at a time

jeromefroe:
This PR adds a `Batch` method to the `index` interface for inserting a batch of documents at a time and adds a corresponding implementation to the in-memory `segment`.


doc/document.go
```diff
@@ -133,35 +136,31 @@ func (d Document) Equal(other Document) bool {
 	return true
 }
 
-// Validate validates the given document and returns its ID if it has one.
-func (d Document) Validate() ([]byte, error) {
+// Validate returns a bool indicating whether the document is valid.
+func (d Document) Validate() error {
```
prateek: should this call `HasID()`? maybe add a test for this case too

index/encoding/doc/doc.go
```diff
@@ -80,8 +80,9 @@ func (w *writer) Open() error {
 }
 
 func (w *writer) Write(d doc.Document) error {
-	w.enc.PutUvarint(uint64(len(d.Fields)))
+	w.enc.PutBytes(d.ID)
```
prateek: The encoder/decoder is beginning to look super similar to the types in https://github.com/m3db/m3db/blob/master/serialize/types.go 

Any chance we can consolidate?

generated/generics/generate.sh
```diff
@@ -32,11 +32,16 @@ fi
 # [1]: https://github.com/cheekybits/genny
 
 mkdir -p $GENERATED_PATH/postingsgen
-cat $GENERIC_MAP_IMPL                                                    \
-  | genny -out=${GENERATED_PATH}/postingsgen/generated_map.go            \
-    -pkg=postingsgen gen "KeyType=[]byte ValueType=postings.MutableList"
+cat $GENERIC_MAP_IMPL                                                  \
+| genny -out=${GENERATED_PATH}/postingsgen/generated_map.go            \
```
prateek: While you're here, could you change the name to `.../postingsgen/map_gen.go` (for this one and others below) . `_gen.go` is the convention Rob started using in m3x/m3db recently. 

And also update `.excludecoverage`

index/types.go
```diff
@@ -31,9 +31,13 @@ import (
...
-	// Insert inserts the given document into the index. The document is guaranteed to be
-	// searchable once the Insert method returns.
-	Insert(d doc.Document) error
+	// Insert inserts the given document into the index and returns its ID. The document
+	// is guaranteed to be searchable once the Insert method returns.
+	Insert(d doc.Document) ([]byte, error)
+
+	// Batch inserts a batch of metrics into the index. The documents are guaranteed to be
+	// searchable all at once when the Batch method returns.
+	Batch(d []doc.Document) error
```
prateek: need something to indicate it's an insert not a read, `InsertBatch` maybe?

index/segment/types.go
```diff
@@ -46,11 +46,11 @@ type Segment interface {
...
+	// Insert inserts the given document into the segment and returns its ID. The document is
+	// guaranteed to be searchable once the Insert method returns.
+	Insert(d doc.Document) ([]byte, error)
 
-	// Seal marks the segment as immutable. After Seal is called no more documents can be
-	// inserted into the segment.
-	Seal() error
+	// Batch inserts a batch of documents into the segment. The documents are guaranteed to
+	// be searchable all at once when the Batch method returns.
+	Batch(d []doc.Document) error
```
prateek: same as below, `InsertBatch`?

index/segment/mem/segment.go
```diff
@@ -22,224 +22,246 @@ package mem
...
-		d.Fields = append(d.Fields, doc.Field{
-			Name:  doc.IDReservedFieldName,
-			Value: uuid,
-		})
-	}
-
-	// TODO: Consider supporting concurrent writes by relaxing the requirement that
-	// inserted documents are immediately searchable.
-	s.ids.Lock()
+	{
```
prateek: hahah nice

index/segment/mem/segment.go
```diff
@@ -22,224 +22,246 @@ package mem
...
+
+			// Update the document since we added an ID.
+			ds[i] = d
+		}
+
+		ids.SetUnsafe(d.ID, struct{}{}, idsgen.SetUnsafeOptions{
+			NoCopyKey:     true,
+			NoFinalizeKey: true,
+		})
+		i++
```
prateek: nit:why not do `i++` in the post completion part of the for loop

prateek: nvm, i see what you're doing in the `ContainsTerm` check. 

Suggestion(take/leave): change the for loop to make it more apparent:

```
i := 0 
for i < len(docs) {
 if  ... {
  continue
 }
i++
}
```

index/segment/mem/segment.go
```diff
@@ -22,224 +22,246 @@ package mem
...
+func (s *segment) insertDocWithLocks(d doc.Document) {
+	nextID := s.writer.nextID
+	s.indexDocWithLock(nextID, d)
+	s.storeDocWithLock(nextID, d)
+	s.writer.nextID++
+}
+
+// indexDocWithLock indexes the fields of a document in the segment's terms dictionary. It
+// must be called with the segment's state lock.
+func (s *segment) indexDocWithLock(id postings.ID, d doc.Document) error {
```
prateek: how bout renaming to `indexDocWithStateRLock` to better indicate what lock you need. 

PS this is why I really want to create https://github.com/m3db/build-tools/issues/14

index/segment/mem/segment.go
```diff
@@ -22,224 +22,246 @@ package mem
...
+	s.termsDict.Insert(doc.Field{
+		Name:  doc.IDReservedFieldName,
+		Value: d.ID,
+	}, id)
+	return nil
+}
+
+// storeDocWithLock stores a documents into the segment's mapping of postings IDs to
+// documents. It must be called with the segment's state lock.
+func (s *segment) storeDocWithLock(id postings.ID, d doc.Document) {
```
prateek: same here, how bout renaming to `storeDocWithStateRLock` to better indicate what lock you need.

doc/document.go
jeromefroe: I don't think we need to require an ID here, in fact, in the segment we currently check if a document is valid before checking its ID. We might want to remove the check for the fields though in case we ever want to index just an ID. Not sure if that would ever be needed though.

index/encoding/doc/doc.go
jeromefroe: I'm not opposed to consolidating though given that they live in separate repos that may be difficult at the moment. Perhaps we can track in an issue?

index/segment/mem/segment.go
jeromefroe: Might be overkill, but I think clearly indicating the scope of the lock visually is helpful :)

index/segment/mem/segment.go
```diff
@@ -22,224 +22,246 @@ package mem
...
 	return nil
 }
 
-func (s *segment) Close() error {
-	s.state.Lock()
-	if s.state.closed {
-		s.state.Unlock()
-		return sgmt.ErrClosed
+func (s *segment) prepareDocs(ds []doc.Document) error {
+	ids := idsgen.New(len(ds))
```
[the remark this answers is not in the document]
jeromefroe: Yea, I wanted to avoid this and push it into the caller but I thought that could break some fundamental assumptions if the client wasn't careful. I also considered moving the `prepareDocs` function into the writer lock so it could be reused. How does that sound as an alternative? I definitely think we're going to need to come back and optimize the segment.

index/encoding/doc/doc.go
prateek: Sure, https://github.com/m3db/m3ninx/issues/36

index/types.go
```diff
@@ -31,9 +31,13 @@ import (
...
-	// Insert inserts the given document into the index. The document is guaranteed to be
-	// searchable once the Insert method returns.
-	Insert(d doc.Document) error
+	// Insert inserts the given document into the index and returns its ID. The document
+	// is guaranteed to be searchable once the Insert method returns.
+	Insert(d doc.Document) ([]byte, error)
+
+	// InsertBatch inserts a batch of metrics into the index. The documents are guaranteed
+	// to be searchable all at once when the Batch method returns.
+	InsertBatch(d []doc.Document) error
```
prateek: Could you indicate which documents were unable to be indexed in the return type

index/segment/mem/segment.go
```diff
@@ -22,224 +22,250 @@ package mem
...
+			if s.termsDict.ContainsTerm(doc.IDReservedFieldName, d.ID) {
+				// The segment already contains this document so we can remove it from those
+				// we need to index.
+				ds[i], ds[len(ds)] = ds[len(ds)], ds[i]
+				ds = ds[:len(ds)-1]
+				continue
+			}
+
+			if _, ok := s.writer.idSet.Get(d.ID); ok {
+				return errDuplicateID
```
prateek: could you make this an exported type

index/segment/mem/segment.go
```diff
@@ -22,224 +22,250 @@ package mem
...
-		return sgmt.ErrClosed
+// prepareDocsWithLocks ensures the given documents can be inserted into the index. It
+// must be called with the state and writer locks.
+func (s *segment) prepareDocsWithLocks(ds []doc.Document) error {
+	s.writer.idSet.Reset()
+	for i := 0; i < len(ds); {
+		d := ds[i]
+		err := d.Validate()
+		if err != nil {
+			return err
```
prateek: instead of early terminating, could you continue attempting to index the remainder of the docs

index/segment/mem/segment.go
```diff
@@ -22,224 +22,250 @@ package mem
...
+		}
+
+		if d.HasID() {
+			if s.termsDict.ContainsTerm(doc.IDReservedFieldName, d.ID) {
+				// The segment already contains this document so we can remove it from those
+				// we need to index.
+				ds[i], ds[len(ds)] = ds[len(ds)], ds[i]
+				ds = ds[:len(ds)-1]
+				continue
```
prateek: could you indicate this case within the return'd type

index/segment/mem/segment.go
prateek: same as above, instead of early terminating, continue to insert other docs

index/types.go
```diff
@@ -42,6 +40,17 @@ type Index interface {
...
 }
 
+// Writer is used to insert documents into an index.
+type Writer interface {
+	// Insert inserts the given document into the index and returns its ID. The document
+	// is guaranteed to be searchable once the Insert method returns.
+	Insert(d doc.Document) ([]byte, error)
+
+	// InsertBatch inserts a batch of metrics into the index. The documents are guaranteed
+	// to be searchable all at once when the Batch method returns.
```
prateek: Maybe document the types of errors this can return so it's not a complete surprise when people are downcasting.

index/segment/mem/segment.go
```diff
@@ -22,224 +22,271 @@ package mem
...
+	{
+		s.writer.Lock()
 
-func (s *segment) Seal() error {
-	s.state.Lock()
-	if s.state.sealed {
-		s.state.Unlock()
-		return errSegmentSealed
+		err = s.prepareDocsWithLocks(b)
+		if err != nil && !index.IsBatchPartialError(err) {
```
prateek: Don't you need to check if partial updates are allowed and insert those still? would be good to add a testcase for this

jeromefroe: This is actually handled in in `prepareDocsWithLocks`. If we don't support partial updates it will return the error right away, otherwise it will continue validating documents and return a partial error (the corresponding test case is `TestSegmentInsertBatchPartialError `)

index/types.go
jeromefroe: Definitely, good idea!

index/segment/mem/segment.go
```diff
@@ -22,224 +22,271 @@ package mem
...
+				b.Docs = b.Docs[:len(b.Docs)-1]
+
+				// Decrement the loop variable since we just removed this document.
+				i--
+				continue
+			}
+
+			if _, ok := s.writer.idSet.Get(d.ID); ok {
+				if !b.AllowPartialUpdates {
+					return errDuplicateID
```
prateek: Could you make `errDuplicateID` an exported type. I want to be able to distinguish it from other failures downstream

index/segment/mem/segment.go
prateek: Cool, good stuff.

PR #37, opened 2018-04-27
[search] Add a negation query

jeromefroe:
This PR adds a negation query.


search/query/conjunction.go
```diff
@@ -28,46 +28,74 @@ import (
 
 // ConjuctionQuery finds documents which match at least one of the given queries.
 type ConjuctionQuery struct {
-	Queries []search.Query
+	Queries   []search.Query
+	Negations []search.Query
```
prateek: cant this be in the `Queries` field? I ask because it'll be annoying to add this to the RPC layer

jeromefroe: Err, crap, yea ideally this would be an internal detail, let's discuss tomorrow

idx/query_matcher.go
```diff
@@ -63,6 +63,8 @@ func (m *queryMatcher) headElem(qry search.Query) (search.Query, bool) {
 		return q, true
 	case *query.RegexpQuery:
 		return q, true
+	case *query.NegationQuery:
```
prateek: probably want to rebase after the other PR lands

idx/query.go
```diff
@@ -51,32 +49,34 @@ func NewRegexpQuery(field, regexp []byte) (Query, error) {
 }
 
 // NewNegationQuery returns a new query for finding documents which don't match a given query.
-func NewNegationQuery(q Query) (Query, error) {
```
prateek: maybe include the error return to be nice to users of the API for now?  We should have the m3ninx+m3db+m3coordinator monorepo sometime next week, then you'll be to change all call-sites more easily.

jeromefroe: I think it would be easier to just rip the band-aid off now. I can update m3db and m3coordinator once this lands.

prateek: sounds good

PR #39, opened 2018-04-28
Add segment iterator

prateek:


postings/range_iterator.go
```diff
@@ -0,0 +1,69 @@
...
+		startInclusive: startInclusive,
+		endExclusive:   endExclusive,
+	}
+}
+
+func (r *rangeIter) Next() bool {
+	if r.closed {
+		return false
+	}
+	if r.started {
```
jeromefroe: nit: would it be simpler to track the next id to avoid needing `started`:

```
func (r *rangeIter) Next() bool {
  if r.closed {
    return false
  }
  r.next++
  return r.next <= r.end
}

func (r *rangeIter) Current() ID {
  return r.next-1
}
```

postings/range_iterator_test.go
```diff
@@ -0,0 +1,46 @@
...
+	require.True(t, iter.Next())
+	require.Equal(t, ID(0), iter.Current())
+	require.True(t, iter.Next())
+	require.Equal(t, ID(1), iter.Current())
+	require.False(t, iter.Next())
+	require.NoError(t, iter.Err())
+	require.NoError(t, iter.Close())
+	require.Error(t, iter.Close())
+}
+func TestRangeIteratorEmpty(t *testing.T) {
```
jeromefroe: nit: add a test for `NewRangeIterator(x, x)` also maybe?

PR #40, opened 2018-04-28
[query] Set number of readers in conjunction and disjunction searchers

jeromefroe:
Previously we set the length of the conjunction and disjunction searchers to be the number of searchers they contained instead of the number of readers the searchers were searching over. The bug was hidden because the test used an equal number of searchers and readers. This diff fixes the bug and ensures the test case uses a different of searchers and readers.

### DIFF
--- a/search/searcher/conjunction.go
+++ b/search/searcher/conjunction.go
@@ -26,8 +26,8 @@ import (
 )
 
 type conjunctionSearcher struct {
-	searchers search.Searchers
-	len       int
+	searchers  search.Searchers
+	numReaders int
 
 	idx  int
 	curr postings.List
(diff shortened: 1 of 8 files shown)


search/searcher/conjunction.go
```diff
@@ -36,14 +36,14 @@ type conjunctionSearcher struct {
...
+func NewConjunctionSearcher(numReaders int, searchers search.Searchers) (search.Searcher, error) {
+	if err := validateSearchers(numReaders, searchers); err != nil {
 		return nil, err
 	}
 
 	return &conjunctionSearcher{
-		searchers: ss,
-		len:       len(ss),
+		searchers: searchers,
+		len:       numReaders,
```
prateek: nit: change the fieldname to `numReaders`

search/searcher/disjunction.go
```diff
@@ -36,14 +36,14 @@ type disjunctionSearcher struct {
...
+func NewDisjunctionSearcher(numReaders int, searchers search.Searchers) (search.Searcher, error) {
+	if err := validateSearchers(numReaders, searchers); err != nil {
 		return nil, err
 	}
 
 	return &disjunctionSearcher{
-		searchers: ss,
-		len:       len(ss),
+		searchers: searchers,
+		len:       numReaders,
```
prateek: same here

PR #46, opened 2018-05-05
Add Segment.Size() method

### LINKED ISSUE #45: Cheap method to retrieve num documents in a Segment
Need a method on `Segment` which returns an estimate for the number of documents it contains. It can be a rough approximation. 

Intent is to use this during a background process in m3db (ticking) to emit metrics ~= cardinality.

prateek:
Resolves #45

### DIFF
--- a/index/segment/mem/segment.go
+++ b/index/segment/mem/segment.go
@@ -82,6 +82,10 @@ func NewSegment(offset postings.ID, opts Options) (sgmt.MutableSegment, error) {
 	return s, nil
 }
 
+func (s *segment) Size() int64 {
+	return int64(int(s.readerID.Load()) - s.offset)
+}
+
 func (s *segment) Insert(d doc.Document) ([]byte, error) {
 	s.state.RLock()
 	defer s.state.RUnlock()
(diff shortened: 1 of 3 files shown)


index/segment/mem/segment.go
```diff
@@ -82,6 +82,10 @@ func NewSegment(offset postings.ID, opts Options) (sgmt.MutableSegment, error) {
 	return s, nil
 }
 
+func (s *segment) Size() int64 {
+	return int64(int(s.readerID.Load()) - s.offset)
```
robskillington: nit: Might be cleaner to `return s.readerID.Load() - int64(s.offset)` to protect against precision (obviously not going to practically be a problem, so take it or leave it).

prateek: Sure thing.

prateek: done

index/segment/mem/segment.go
```diff
@@ -82,6 +82,10 @@ func NewSegment(offset postings.ID, opts Options) (sgmt.MutableSegment, error) {
 	return s, nil
 }
 
+func (s *segment) Size() int64 {
+	return int64(s.readerID.Load()) - int64(s.offset)
```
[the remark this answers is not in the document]
prateek: talked offline: don't strictly need to, but better to do it for consistency across methods. addressed in https://github.com/m3db/m3ninx/pull/50

PR #48, opened 2018-05-05
Add Segment.ContainsID(id)

prateek:


index/segment/mem/segment.go
```diff
@@ -82,6 +82,20 @@ func NewSegment(offset postings.ID, opts Options) (sgmt.MutableSegment, error) {
 	return s, nil
 }
 
+func (s *segment) ContainsID(id []byte) (bool, error) {
+	s.state.RLock()
+	defer s.state.RUnlock()
```
robskillington: nit: This might be on the hot path, so avoid defer *may* be ideal.

PR #51, opened 2018-05-09
Allow docs w/ no fields if they have IDs

prateek:


doc/document.go
```diff
@@ -138,7 +138,7 @@ func (d Document) Equal(other Document) bool {
 
 // Validate returns a bool indicating whether the document is valid.
 func (d Document) Validate() error {
-	if len(d.Fields) == 0 {
+	if len(d.Fields) == 0 && !d.HasID() {
```
jeromefroe: maybe just a test for this case if its easy? I have to-do to go back and improve test coverage soon so can also get it then

prateek: added

PR #54, opened 2018-05-09
[matcher] Simplify matching by using reflect

jeromefroe:


prateek:
Any reason to avoid using the `Equal()` methods?

jeromefroe:
Just hadn't implemented `Equal` for the structs in question, can definitely add it if we need it more generally.

PR #56, opened 2018-05-10
[query] Add methods to serialize queries

jeromefroe:
Add methods to serialize queries.


idx/codec.go
```diff
@@ -0,0 +1,39 @@
...
+// THE SOFTWARE.
+
+package idx
+
+import (
+	"github.com/m3db/m3ninx/search/query"
+)
+
+// Marshal encodes the query into a byte slice.
+func Marshal(q Query) ([]byte, error) {
```
prateek: +1 for containing the proto dep outside the `idx` package.

PR #60, opened 2018-05-12
Add access to fieldnames/values/postingslists

prateek:
- Add accessors to segment to prepare for immutable segment
- Codecov excludes for generated files
- mock-gen'd updates
- update m3x/roaring versions

### DIFF
--- a/index/segment/mem/segment.go
+++ b/index/segment/mem/segment.go
@@ -32,10 +32,14 @@ import (
 	"github.com/m3db/m3ninx/postings"
 )
 
-// ErrDuplicateID is the error returned when a batch contains duplicate IDs.
-var ErrDuplicateID = errors.New("a batch cannot contain duplicate IDs")
+var (
+	// ErrDuplicateID is the error returned when a batch contains duplicate IDs.
+	ErrDuplicateID = errors.New("a batch cannot contain duplicate IDs")
 
-var errUnknownPostingsID = errors.New("unknown postings ID specified")
+	errUnknownPostingsID = errors.New("unknown postings ID specified")
+	errSegmentSealed     = errors.New("unable to seal, segment has already been sealed")
+	errSegmentIsUnsealed = errors.New("un-supported operation on an un-sealed mutable segment")
+)
 
 // nolint: maligned
 type segment struct {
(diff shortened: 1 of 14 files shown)


index/segment/mem/segment.go
```diff
@@ -369,3 +374,54 @@ func (s *segment) Close() error {
 	s.state.closed = true
 	return nil
 }
+
+func (s *segment) Seal() (sgmt.Segment, error) {
+	s.state.Lock()
+	defer s.state.Unlock()
+	if s.state.closed {
```
jeromefroe: nit: maybe it would be worth moving these checks into `checkIsOpenWithLock` `checkIsSealedWithLock` functions?

index/segment/mem/segment_test.go
```diff
@@ -754,3 +801,45 @@ func compareDocs(expected, actual doc.Document) bool {
 	}
 	return expected.Equal(actual)
 }
+
+var (
```
jeromefroe: nit: move these to the top of the file?

index/segment/mem/segment.go
prateek: sure

prateek: actually, I think this hurts readability more than helping it - 
screen shot 2018-05-12 at 11 55 53

Going to leave as is.

jeromefroe: Ahh, so what I meant was not have a check for sealed or closed but to coalesce them, i.e.

```
func (s * segment) checkIsOpenWithLock() bool {
  return !s.state.closed && ! s.state.sealed
}
```

index/segment/mem/segment_test.go
prateek: sure thing

index/segment/mem/segment.go
prateek: meh i can if you feel strongly but i think having the separate errors returned is also kinda nice.

jeromefroe: don't feel strongly, but you could also keep the separate errors:

```
func (s * segment) checkIsOpenWithLock() error {
  if s.state.closed {
    return errClosed
  }
  if s.state.sealed {
    return errSealed
  }
  return nil
}
```

prateek: Sure

prateek:
Wrong base branch. Sigh.

prateek:
fixed in #61

PR #62, opened 2018-05-12
[fs] Refactor stored fields file into data and index files

jeromefroe:
This PR refactors the stored fields file into two separate files: a data file containing the stored fields for each document in a segment and an index file which maps a postings ID to its offset in the data file.


index/segment/fs/encoding/encoding.go
```diff
@@ -62,16 +65,19 @@ func (e *Encoder) PutUint64(x uint64) {
 	e.buf = append(e.buf, e.tmp[:8]...)
 }
 
-// PutUvarint encodes a variable-sized unsigned integer.
-func (e *Encoder) PutUvarint(x uint64) {
+// PutUvarint encodes a variable-sized unsigned integer and returns the number of
+// bytes written.
+func (e *Encoder) PutUvarint(x uint64) int {
```
prateek: nit: make PutUint32 and other methods return the int too .

index/segment/fs/encoding/fields/README.md
```diff
@@ -0,0 +1,128 @@
...
+│ │      Field Value      │ │
+│ │        (bytes)        │ │
+│ │                       │ │
+│ └───────────────────────┘ │
+└───────────────────────────┘
+```
+
+## Index File
+
+The index file contains, for each postings ID in the segment, the offset of the corresponding
```
prateek: could you add the caveat that offsets set to maxuint64 or w/e indicates the doc doesn't exist

index/segment/fs/encoding/fields/index.go
```diff
@@ -0,0 +1,145 @@
...
+	r.limit = r.base + postings.ID(count)
+	return r, nil
+}
+
+func (r *IndexReader) Read(id postings.ID) (uint64, error) {
+	if id < r.base || id > r.limit {
+		return 0, fmt.Errorf("invalid postings ID %v, must be between [%v, %v]", id, r.base, r.limit)
+	}
+
+	idx := (int(id-r.base) * 8) + indexHeaderSize
```
prateek: hm don't you need + 8 here for the base id field?

prateek: i.e.
```
idx : = (int(id-r.base) * 8) + indexHeaderSize + 8
```

prateek: would be good to extract out the id -> offset -> id translation into helper methods so you don't have to chase the assumptions everywhere in the codebase.

index/segment/fs/encoding/fields/index.go
```diff
@@ -0,0 +1,145 @@
...
+func NewIndexReader(data []byte) (*IndexReader, error) {
+	if len(data) < indexHeaderSize {
+		return nil, io.ErrShortBuffer
+	}
+
+	payloadLen := len(data) - indexHeaderSize
+	if payloadLen%8 != 0 {
+		return nil, fmt.Errorf("stored fields index payload should be a multiple of 8, found %v", payloadLen%8)
+	}
+	count := payloadLen / 8
```
prateek: hm, don't you need to remove the base id from these calcs?

index/segment/fs/encoding/fields/data.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (w *DataWriter) Write(d doc.Document) (int, error) {
+	n := w.enc.PutBytes(d.ID)
+	n += w.enc.PutUvarint(uint64(len(d.Fields)))
+	for _, f := range d.Fields {
+		n += w.enc.PutBytes(f.Name)
+		n += w.enc.PutBytes(f.Value)
+	}
+
+	if w.enc.Len() > dataEncoderWriteThreshold {
```
prateek: same as the index writer, why not rely on the external callers using buffered writers if they want this behaviour?

index/segment/fs/encoding/fields/data.go
```diff
@@ -0,0 +1,140 @@
...
+// NewDataReader returns a new DataReader.
+func NewDataReader(data []byte) *DataReader {
+	return &DataReader{
+		data: data,
+		dec:  encoding.NewDecoder(nil),
+	}
+}
+
+func (r *DataReader) Read(offset uint64) (doc.Document, error) {
+	if offset > uint64(len(r.data)) {
```
prateek: should this be `>=` ?

index/segment/fs/encoding/fields/README.md
jeromefroe: I'd prefer to leave that out until we actually need it. Seems like an unnecessary complexity at this point and it's easy to add in the future.

index/segment/fs/encoding/fields/data.go
jeromefroe: That's fair, my concern was that if we were given a file handle, this could lead to a lot of syscalls for small writes. Push that decision to the caller sounds good to me though.

index/segment/fs/encoding/fields/index.go
```diff
@@ -0,0 +1,145 @@
...
+type IndexReader struct {
+	data  []byte
+	dec   *encoding.Decoder
+	base  postings.ID
+	limit postings.ID
+}
+
+// NewIndexReader returns a new IndexReader.
+func NewIndexReader(data []byte) (*IndexReader, error) {
+	if len(data) < indexHeaderSize {
```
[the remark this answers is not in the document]
jeromefroe: It writes out in the first call to `Write` (line 56).

index/segment/fs/encoding/fields/index.go
jeromefroe: That's done in line 112 when we subtract `indexHeaderSize `, right now the header only contains the base ID.

index/segment/fs/encoding/fields/index.go
jeromefroe: The header is the base ID field. Moving into a helper function sounds like a good idea.

index/segment/fs/encoding/fields/index.go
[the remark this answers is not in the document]
prateek: I see. Maybe don't call it header if you're going to encode the base id there. I thought you were using a const header with some other metadata.

index/segment/fs/encoding/fields/index.go
prateek: Yeah the var name confused me. maybe change to indicate it's the base id?

index/segment/fs/encoding/fields/README.md
jeromefroe: Talked offline, will add.

index/segment/fs/encoding/fields/index.go
```diff
@@ -0,0 +1,145 @@
...
+// NewIndexWriter returns a new IndexWriter.
+func NewIndexWriter(w io.Writer) *IndexWriter {
+	iw := &IndexWriter{
+		writer: w,
+		enc:    encoding.NewEncoder(initialIndexEncoderLen),
+	}
+	return iw
+}
+
+func (w *IndexWriter) Write(id postings.ID, offset uint64) error {
```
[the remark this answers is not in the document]
jeromefroe: Sure thing, will add a comment.

index/segment/fs/encoding/fields/index.go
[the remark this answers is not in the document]
jeromefroe: s/header/metadata?

index/segment/fs/encoding/fields/index.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not read base postings ID: %v", err)
+	}
+	r.base = postings.ID(base)
+	r.limit = r.base + postings.ID(count)
+	return r, nil
+}
+
+func (r *IndexReader) Read(id postings.ID) (uint64, error) {
+	if id < r.base || id >= r.limit {
```
prateek: i think this needs to be `> r.limit`

jeromefroe: The limit is the base plus the length so I think this should be `>=` much like when comparing to the length of a slice. The tests ensure they get all ids we set so I think we're ok.

PR #64, opened 2018-05-13
Persistent Segment using FSTs

prateek:
pending for this pr:
- [x] quick readme detailing the different file formats and internals 

next steps:
- [ ] merge w/ jerome's changes for document data & index file writer, and more e2e testing

followups:
- [ ] heavier weight integration tests that interact with an actual fs
- [ ] prop test suite
- [ ] rudimentary fuzz tests to ensure we're not going to segfault on bad data


generated-source-files.mk
```diff
@@ -52,8 +52,8 @@ genny-map-segment-mem-postingsmap: install-m3x-repo
 		key_type=[]byte                                    \
 		value_type=postings.MutableList                    \
 		target_package=$(m3ninx_package)/index/segment/mem \
-		rename_nogen_key=forsure                           \
-		rename_nogen_value=forsure                         \
+		rename_nogen_key=yes                           \
```
prateek: s/yes/true

index/segment/fst/roaring_postings_codec.go
```diff
@@ -0,0 +1,63 @@
+// +build disabled
+// TODO(prateek): measure perf of this v pilosa
```
prateek: rudimentary tests show its 2x worse. should drop it.

index/segment/tests/fst_mem_simple_test.go
```diff
@@ -0,0 +1,283 @@
...
+func newTestSegments(t *testing.T, docs []doc.Document) (memSeg segment.MutableSegment, fstSeg segment.Segment) {
+	s := newTestMemSegment(t)
+	for _, d := range docs {
+		_, err := s.Insert(d)
+		require.NoError(t, err)
+	}
+	return s, newFSTSegment(t, s)
+}
+
+// TODO(prateek): follow up if this is a fair requirement in everyone's mind.
```
prateek: note to self: have to track some issues in the implementation for 0 length field-values e.g. `"xyz"` => `""`. It's somewhere in fst read path.

@robskillington Do you think it's a fair requirement to require all tags to have non-empty name/value?

postings/types.go
```diff
@@ -69,6 +70,9 @@ type List interface {
 
 	// Equal returns whether this postings list contains the same posting IDs as other.
 	Equal(other List) bool
+
+	// WriteTo writes the postings list to the provided writer.
+	WriteTo(io.Writer) (int64, error)
```
prateek: will drop this method

postings/types.go
```diff
@@ -90,6 +94,12 @@ type MutableList interface {
 	// this postings list or other.
 	Union(other List) error
 
+	// AddIterator adds all IDs contained in the iterator.
+	AddIterator(iter Iterator) error
+
+	// ReadFrom reads a postings list from the provided reader.
+	ReadFrom(io.Reader) (int64, error)
```
prateek: this one too

postings/types.go
```diff
@@ -90,6 +94,12 @@ type MutableList interface {
 	// this postings list or other.
 	Union(other List) error
 
+	// AddIterator adds all IDs contained in the iterator.
+	AddIterator(iter Iterator) error
```
prateek: will retain this one

.excludemetalint
```diff
@@ -3,3 +3,4 @@ generated/
 _mock.go
 .pb.go
 index/segment/mem/ids_map_gen.go
+index/segment/fst/roaring_postings_codec.go
```
prateek: undo this once you drop the file

index/segment/fst/pilosa_postings_codec.go
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
prateek: drop this file

index/segment/fst/pilosa_postings_codec.go
```diff
@@ -0,0 +1,123 @@
...
+	if err := b.UnmarshalBinary(data); err != nil {
+		return nil, err
+	}
+
+	pl := pool.Get()
+	iter := newIter(b.Iterator())
+	return pl, pl.AddIterator(iter)
+}
+
+// NB: need to do this to find a path into our postings list which doesn't require every
```
prateek: open followup issue for this

index/segment/fst/reader.go
```diff
@@ -0,0 +1,354 @@
...
+}
+
+func (r *reader) retrievePostingsList(postingsOffset uint64) (postings.List, error) {
+	postingsBytes, err := r.retrieveBytes(r.opts.PostingsData, postingsOffset)
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve postings data: %v", err)
+	}
+
+	return fromPilosa(postingsBytes, r.opts.PostingsListPool)
+	// return fromRoaringBytes(postingsBytes, r.opts.PostingsListPool)
```
prateek: delete

index/segment/fst/reader.go
```diff
@@ -0,0 +1,354 @@
...
+	r.Lock()
+	defer r.Unlock()
+	if r.closed {
+		return errReaderClosed
+	}
+	r.closed = true
+	return r.fieldsFST.Close()
+}
+
+// FOLLOWUP(prateek): really need to change the types returned in Fields() and Terms()
```
prateek: file ticket

index/segment/fst/roaring_postings_codec.go
robskillington: glad to hear the pilosa library is quite good 👍, definitely interested to hear about what you observed at some point 📉

index/segment/tests/fst_mem_simple_test.go
robskillington: Hm yeah that's probably fair, the question is though, where can/should we block this.  Potentially at the `db.WriteTagged` method we can do the validation there and bubble back an `InvalidParamsError` if so.

glide.yaml
```diff
@@ -24,6 +23,14 @@ import:
   version: ^1.0.0
   subpackages:
   - proto
+- package: github.com/couchbase/vellum
+  version: eb6ae3743b3f300f2136f83ca78c08cc071edbd4
+- package: github.com/edsrzf/mmap-go # un-used but required for a compile time dep from vellum
+  version: 0bce6a6887123b67a60366d2c9fe2dfb74289d2e
+- package: github.com/pilosa/pilosa
```
robskillington: Might it be worth specifying that you're only pulling in `github.com/pilosa/pilosa/roaring`?

index/segment/fst/fst_codec.go
```diff
@@ -0,0 +1,65 @@
...
+	builder, err := vellum.New(&f.scratchBuffer, nil)
+	if err != nil {
+		return err
+	}
+	f.builder = builder
+	return nil
+}
+
+func (f *fstCodec) Add(b []byte, v uint64) error {
+	return f.builder.Insert(b, v)
```
robskillington: Does `vellum.Builder` check if insert was in order or not?  If not should this codec perform that check by comparing against last insert?

index/segment/fst/fst_codec.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (f *fstCodec) Bytes() ([]byte, error) {
+	if err := f.builder.Close(); err != nil {
```
robskillington: I suppose because of this you can't call `Bytes()` twice.  May it be worth calling this instead `Done()` or something instead of `Bytes()`?

index/segment/fst/pilosa_postings_codec.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func newIter(iter *pilosa.Iterator) *pilosaBitmapIterator {
+	return &pilosaBitmapIterator{
+		iter: iter,
+	}
+}
+
+func (p *pilosaBitmapIterator) Next() bool {
+	v, eof := p.iter.Next()
```
robskillington: Does this give EOF just once?  If so should the EOF be cached, perhaps `if !p.hasNext { return false }` early from this method?

index/segment/fst/reader.go
```diff
@@ -0,0 +1,354 @@
...
+	termsFST, err := r.retrieveTermsFST(field)
+	if err != nil {
+		return nil, err
+	}
+	defer termsFST.Close()
+
+	pl := r.opts.PostingsListPool.Get()
+
+	iter, iterErr := termsFST.Search(re, minByteKey, maxByteKey)
+	defer iter.Close()
```
robskillington: nit: Combine the defers?

robskillington: Also do you need to check return type of `iter.Close()` here?

index/segment/fst/reader.go
```diff
@@ -0,0 +1,354 @@
...
+	const sizeofUint64 = 8
+	var (
+		sizeStart = offset
+		sizeEnd   = sizeStart + sizeofUint64
+	)
+	if sizeEnd > uint64(len(base)) {
+		return nil, fmt.Errorf("base bytes too small, length: %d, base-offset: %d", len(base), offset)
+	}
+
+	sizeBytes := base[offset : offset+sizeofUint64]
```
robskillington: nit: This can be `base[sizeStart:sizeEnd]`?

robskillington: This can be `base[sizeStart:sizeEnd` yeah?

index/segment/fst/types.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/m3db/m3ninx/postings"
+)
+
+const (
+	majorVersion = 1
+	minorVersion = 0
+)
+
+// Reader represents a FST segment.
+type Reader interface {
```
robskillington: Hm, is there a reason you don't embed `index.Reader` and instead call out the methods it implements explicitly?

index/segment/fst/roaring_postings_codec.go
prateek: For sure. So using the tests as a poor man's benchmark:

```
# pilosa 
❯ time go test ./index/segment/tests -count=10
ok  	github.com/m3db/m3ninx/index/segment/tests 2.192s
/Users/prungta/code/go1.10.2/bin/go test ./index/segment/tests -count=10  3.81s user 0.47s system 131% cpu 3.266 total

# roaring binary 
❯ time go test ./index/segment/tests -count=10
ok  	github.com/m3db/m3ninx/index/segment/tests 1.966s
/Users/prungta/code/go1.10.2/bin/go test ./index/segment/tests -count=10  3.76s user 0.46s system 137% cpu 3.073 total


# roaring msgpack 
❯ time go test ./index/segment/tests -count=10
ok  	github.com/m3db/m3ninx/index/segment/test 8.219s
/Users/prungta/code/go1.10.2/bin/go test ./index/segment/tests -count=10  17.38s user 0.78s system 197% cpu 9.211 total
```

conclusions:
- upstream roaring msgpack is a pos
- binary perf of upstream roaring v pilosa is comparable. 

This is despite the fact that in the case of the pilosa tests, we're only using pilosa for serialization; and always convert from []byte -> pilosa -> upstream roaring (and vice versa during writes) to satisfy existing interfaces. So once we switch our postings implementation to pilosa, it'd be able to avoid a bunch of things it's having to do now. 

Also checked mem usage, binary roaring v pilosa were comparable - 94MB v 105MB alloc space (again, despite the extra work our pilosa impl had to do for conversions)
roaring-binary-mem
pilosa-mem

.excludemetalint
prateek: done

generated-source-files.mk
prateek: done

glide.yaml
prateek: hm doesn't the sub-package specifier below this indicate that?

index/segment/fst/fst_codec.go
prateek: Yep vellum does: https://github.com/couchbase/vellum/blob/master/builder.go#L94-L96

index/segment/fst/fst_codec.go
prateek: Yep so reworked the fstCodec into an fstWriter to never directly buffer fst data in memory (vellum will buffer what it needs, but we don't anymore). 

Ended up renaming this method to `Close()`

index/segment/fst/pilosa_postings_codec.go
prateek: fair will do

index/segment/fst/reader.go
prateek: done

index/segment/fst/reader.go
prateek: done

postings/types.go
prateek: done

postings/types.go
prateek: done

index/segment/fst/types.go
prateek: lol it's cause this type shares the same name :| Can't have interface X embed another interface X even if it's another package. Will rename this to Readable to avoid this issue.

prateek: actually can't do that, cause both Reader and Segment have Close() which conflicts when embedding.

index/segment/fst/reader.go
prateek: sure thing, started using the SafeCloser idiom here

prateek: done

index/segment/fst/reader.go
```diff
@@ -0,0 +1,354 @@
...
+func (r *reader) Docs(pl postings.List) (doc.Iterator, error) {
+	return nil, errNotImplemented
+}
+
+func (r *reader) AllDocs() (doc.Iterator, error) {
+	return nil, errNotImplemented
+}
+
+func (r *reader) retrievePostingsList(postingsOffset uint64) (postings.List, error) {
+	postingsBytes, err := r.retrieveBytes(r.opts.PostingsData, postingsOffset)
```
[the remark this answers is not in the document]
prateek: Yeah exactly. it's either copy on read or tie the lifecycles together correctly. Much easier to do the copy on read and pool backing slices IMO.

index/segment/fst/reader.go
prateek: lol yep. made the change.

index/segment/fst/reader.go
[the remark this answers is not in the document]
prateek: Not pooling right now: will need to return iterator types everywhere to make that easier on API users.

index/segment/tests/fst_mem_simple_test.go
prateek: Definitely. So plan for m3db would be to do that at Client/WriteTagged on client side, and in serialize.TagDecoder on server side. i.e. we'd have to wait until we're sure we need to decode the tags so that we don't do this excessively. 

For m3ninx, it'd be in `doc.Validate()` /cc @jeromefroe

index/segment/fst/reader.go
```diff
@@ -0,0 +1,469 @@
...
+	}
+
+	termsFST, err := r.retrieveTermsFST(field)
+	if err != nil {
+		return nil, err
+	}
+
+	terms, err := r.allKeys(termsFST)
+	if err != nil {
+		termsFST.Close()
```
robskillington: Use safe closer for this too?

index/segment/fst/reader.go
```diff
@@ -0,0 +1,469 @@
...
+// where size/magicNumber are strictly uint64 (i.e. 8 bytes). It assumes the 8 bytes preceding the offset
+// are the magicNumber, the 8 bytes before that are the size, and the `size` bytes before that are the
+// payload. It retrieves the payload while doing bounds checks to ensure no segfaults.
+func (r *reader) retrieveBytes(base []byte, offset uint64) ([]byte, error) {
+	const sizeofUint64 = 8
+	var (
+		magicNumberEnd   = int64(offset) // to prevent underflows
+		magicNumberStart = offset - sizeofUint64
+	)
+	if magicNumberEnd > int64(len(base)) || magicNumberStart < 0 {
```
robskillington: Shouldn't this be `>=` i.e. `magicNumberEnd >= int64(len(base))`?

index/segment/fst/types.go
robskillington: Ugh.  Can you do something hacky like below to at least enforce it always meets the `index.Reader`?

index/segment/fst/writer.go
```diff
@@ -0,0 +1,361 @@
...
+		fstTermsOffsets: newFSTTermsOffsetsMap(defaultInitialFSTTermsOffsetsMapSize),
+	}
+}
+
+func (w *writer) Clear() {
+	w.seg = nil
+
+	w.intEncoder.Reset()
+	w.postingsEncoder.Reset()
+	w.fstWriter.Clear()
```
robskillington: Any reason why fstWriter is Clear not Reset?

index/segment/fst/writer.go
```diff
@@ -0,0 +1,361 @@
...
+
+func (w *writer) WriteDocumentsIndex(iow io.Writer) error {
+	if !w.docsDataFileWritten {
+		return fmt.Errorf("documents data file has to be written before documents index file")
+	}
+	return nil
+}
+
+func (w *writer) WritePostingsOffsets(iow io.Writer) error {
+	currentOffset := uint64(0)
```
robskillington: Shouldn't this also be `if !w.docsDataFileWritten || !w.docsDataIndexFileWritten { ... }`?  Can stub out in tests by forcing them to be true for now I guess if it was a pain to test them out.

index/segment/fst/writer.go
```diff
@@ -0,0 +1,361 @@
...
+			w.addPostingsOffset(currentOffset, f, t)
+		}
+	}
+
+	w.postingsFileWritten = true
+	return nil
+}
+
+func (w *writer) WriteFSTTerms(iow io.Writer) error {
+	if !w.postingsFileWritten {
```
robskillington: Should also test the other files written too?

index/segment/fst/writer.go
```diff
@@ -0,0 +1,361 @@
...
+	numBytesWritten += uint64(size)
+	n, err := w.writeSizeMagicNumber(iow, uint64(size))
+	if err != nil {
+		return 0, err
+	}
+	numBytesWritten += n
+	return numBytesWritten, nil
+}
+
+func (w *writer) writeSizeMagicNumber(iow io.Writer, size uint64) (uint64, error) {
```
robskillington: nit: writeSizeAndMagicNumber?  I wasn't sure if size was a magic number or something on first read.

generated/proto/fstwriter/fstwriter.proto
```diff
@@ -0,0 +1,45 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+syntax = "proto3";
+package fstwriter;
+
+option go_package = "fstwriter";
```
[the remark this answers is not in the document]
prateek: fair will remove

index/segment/fst/reader.go
prateek: sure thing.

index/segment/fst/writer.go
prateek: It has Reset() too, but that requires an argument (io.Writer). It's Clear() releases any held resources, and Reset() makes it ready to write out data.

index/segment/fst/writer.go
prateek: Nah, so the identifiers in the postings data is `doc.ID`, which doesn't depend on where/how the documents data is persisted. So don't need the ordering constraint here.

index/segment/fst/writer.go
prateek: We can rely on the check being correct transitively yea? As long as each file checks it's requirement, we get the same total guarantee.

index/segment/fst/writer.go
prateek: sure thing

prateek: done

index/segment/fst/reader.go
prateek: The offset we persist is past the `(payload, size, magicNumber)` boundary, and because we 0-index the underlying []byte, `offset = len(base)` is the last valid offset. 

If it makes any you feel any better, the current tests break if I make the change.

generated/proto/fstwriter/fstwriter.proto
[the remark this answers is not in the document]
prateek: done

index/segment/fst/types.go
prateek: done

index/segment/fst/reader.go
prateek: done

index/segment/fst/README.md
```diff
@@ -0,0 +1,41 @@
+FST Segment
```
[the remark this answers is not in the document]
prateek: Cheers

index/segment/fst/reader.go
```diff
@@ -0,0 +1,471 @@
...
+
+	if err := fstCloser.Close(); err != nil {
+		return nil, err
+	}
+
+	return pl, nil
+}
+
+func (r *reader) MatchRegexp(field []byte, regexp []byte, compiled *regexp.Regexp) (postings.List, error) {
+	re, err := vregex.New(string(regexp))
```
jeromefroe: do we need to check if the reader is closed here?

index/segment/fst/README.md
```diff
@@ -0,0 +1,45 @@
...
+===========
+
+```
+
+┌───────────────────────────────┐           ┌───────────────────────────────┐
```
jeromefroe: Any reason to not use a single FST for the fields and terms using a non-UTF8 byte as the separator?

index/segment/fst/reader.go
```diff
@@ -0,0 +1,471 @@
...
+
+func (r *reader) Docs(pl postings.List) (doc.Iterator, error) {
+	return nil, errNotImplemented
+}
+
+func (r *reader) AllDocs() (doc.Iterator, error) {
+	return nil, errNotImplemented
+}
+
+func (r *reader) retrievePostingsList(postingsOffset uint64) (postings.List, error) {
```
jeromefroe: nit: is it worth signaling that this must be called the `RLock` in the name? Same with those below?

index/segment/fst/types.go
jeromefroe: What about calling it a `ReaderSegment` or `ReadableSegment`?

index/segment/fst/types.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// Writer writes out a FST segment from the provided elements.
+type Writer interface {
+	// Clear clears the internal state of the Writer to allow re-use.
+	Clear()
+
+	// Reset sets the Writer to persist the provide segment.
+	// NB(prateek): the provided segment must be a Sealed Mutable segment.
+	Reset(s segment.MutableSegment) error
```
jeromefroe: What about `Init`? I'm not sure what the difference is between `Clear` and `Reset`?

index/util/docs.go
```diff
@@ -71,7 +71,24 @@ func ReadDocs(path string, n int) ([]doc.Document, error) {
 		return nil, fmt.Errorf("requested %d metrics but found %d", n, len(docs))
 	}
 
-	return docs, nil
+	return filterEmptyFields(docs), nil
+}
+
+func filterEmptyFields(docs []doc.Document) []doc.Document {
```
jeromefroe: Would it be worth moving this logic into `ReadDocs` so we can just remove empty fields instead of dropping the whole document? Is this a common occurrence?

index/segment/fst/reader.go
```diff
@@ -0,0 +1,471 @@
...
+	if magicNumberEnd > int64(len(base)) || magicNumberStart < 0 {
+		return nil, fmt.Errorf("base bytes too small, length: %d, base-offset: %d", len(base), magicNumberEnd)
+	}
+	magicNumberBytes := base[magicNumberStart:magicNumberEnd]
+	d := encoding.NewDecoder(magicNumberBytes)
+	n, err := d.Uint64()
+	if err != nil {
+		return nil, fmt.Errorf("error while decoding magicNumber: %v", err)
+	}
+	if n != uint64(magicNumber) {
```
jeromefroe: Why include a magic number in each tuple?

index/segment/fst/types.go
jeromefroe: Talked offline, move `Clear` to `Reset(nil)`

generated/proto/fstwriter/fstwriter.proto
```diff
@@ -0,0 +1,43 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+syntax = "proto3";
+package fstwriter;
```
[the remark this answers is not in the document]
prateek: done

generated/proto/fstwriter/fstwriter.proto
```diff
@@ -0,0 +1,43 @@
...
+
+enum SegmentType {
```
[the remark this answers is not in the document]
prateek: talked offline, leaving as is for now.

index/util/docs.go
prateek: done

prateek: done

index/segment/fst/reader.go
```diff
@@ -0,0 +1,471 @@
...
+	if r.closed {
+		return false, errReaderClosed
+	}
+
+	termsFST, err := r.retrieveTermsFST(doc.IDReservedFieldName)
+	if err != nil {
+		return false, err
+	}
+	fstCloser := x.NewSafeCloser(termsFST)
+	defer fstCloser.Close()
```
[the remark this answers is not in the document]
prateek: talked offline, leaving as is.

index/segment/fst/README.md
prateek: talked offline: vellum's regexp parser doesn't support non-utf8 chars. example here -https://gist.github.com/prateek/fb2cb1a197d4dcfeaf39640a40af3787

index/segment/fst/reader.go
prateek: done

index/segment/fst/reader.go
prateek: talked offline: used to delineate object boundaries.

index/segment/fs/segment.go
```diff
@@ -0,0 +1,477 @@
...
+	errReaderClosed            = errors.New("segment is closed")
+	errUnsupportedMajorVersion = errors.New("unsupported major version")
+	errNotImplemented          = errors.New("operation not implemented")
+
+	minByteKey = []byte{}
+	maxByteKey = []byte(string(utf8.MaxRune))
+)
+
+// NewSegmentOpts represent the collection of parameters to construct a Segment.
+type NewSegmentOpts struct {
```
jeromefroe: nit: maybe `SegmentData` would be more appropriate since these values aren't actually optional?

index/segment/fs/segment.go
```diff
@@ -0,0 +1,477 @@
...
+}
+
+type segmentReader struct {
+	sync.RWMutex
+	closed bool
+
+	segment *segment
+}
+
+var _ index.Reader = &segmentReader{}
```
jeromefroe: unneeded now?

index/segment/fs/segment.go
```diff
@@ -0,0 +1,477 @@
...
+		opts:    opts,
+		numDocs: metadata.NumDocs,
+	}, nil
+}
+
+type segment struct {
+	sync.RWMutex
+	closed bool
+
+	fieldsFST *vellum.FST
```
jeromefroe: Would it be worth just using a sorted list for the fields instead of an FST since we don't need to search over the fields?

index/segment/fs/segment.go
```diff
@@ -0,0 +1,477 @@
...
+	}
+
+	if err := safeCloser.Close(); err != nil {
+		return nil, err
+	}
+	return keys, nil
+}
+
+func (r *segment) retrieveTermsFSTWithRLock(field []byte) (*vellum.FST, error) {
+	termsFSTOffset, exists, err := r.fieldsFST.Get(field)
```
jeromefroe: Would it be worth caching the terms FSTs in a map from offset to FST so we don't have to reload them each time?

postings/pilosa/codec.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = pilosaBitmap.WriteTo(&e.scratchBuffer)
+	if err != nil {
+		return nil, err
+	}
+
+	return e.scratchBuffer.Bytes(), nil
```
jeromefroe: Would we need to return a copy of these bytes since we reuse the buffer?

postings/pilosa/codec.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// Reset resets the internal state of the encoder to allow
+// for re-use.
+func (e *Encoder) Reset() {
+	e.scratchBuffer.Reset()
+}
+
+// Encode encodes the provided postings list.
+func (e *Encoder) Encode(pl postings.List) ([]byte, error) {
```
jeromefroe: nit: Thoughts on calling it `Marshal` since we have `Unmarshal` below?

postings/pilosa/iterator.go
```diff
@@ -0,0 +1,71 @@
...
+// insert to grab a lock. Need to make a non thread-safe version of our api.
+// FOLLOWUP(prateek): tracking this issue in https://github.com/m3db/m3ninx/issues/65
+
+type iterator struct {
+	iter    *roaring.Iterator
+	current uint64
+	hasNext bool
+}
+
+var _ postings.Iterator = &iterator{}
```
jeromefroe: Is this covered by the constructor below?

x/safe_closer.go
```diff
@@ -0,0 +1,50 @@
...
+func (c *safeCloser) Close() error {
+	if c.closed {
+		return nil
+	}
+	c.closed = true
+	return c.Closable.Close()
+}
+
+// Closable represents a closable resource.
+type Closable interface {
```
jeromefroe: Worth just using the standard library's `io.Closer`?

index/segment/fs/segment.go
prateek: sure, i'll break into two structs for this.

index/segment/fs/segment.go
prateek: still needed. they're separate types

index/segment/fs/segment.go
prateek: No. A FST will compress much more than the sorted list, and will be faster than a sorted list at finding a field. Further we don't search over fields today, but no reason we can't support it in the future.

index/segment/fs/segment.go
prateek: Yes, can be a subsequent PR based on perf work we have do to for query anyway?

postings/pilosa/codec.go
prateek: Can push the responsibility on the caller. Added a note about sematics to the method comment.

postings/pilosa/codec.go
prateek: I'd do that if it was a function on the package not the struct. I'm going to leave as is for now.

postings/pilosa/iterator.go
prateek: It is, but i prefer specifying it like this in addition to capture the intent of type separate from the constructor.

x/safe_closer.go
prateek: Sure

prateek: done

index/segment/fs/segment.go
jeromefroe: I agree about compression, but I think in terms of performance it would be inferior. Can perhaps perform the benchmarks some time especially if we want to consider any B-Tree libraries.

postings/pilosa/codec.go
jeromefroe: What about having the caller provide the buffer so it's explicit that they control the lifetime of the buffer?

index/segment/fs/segment.go
robskillington: It'd have to be file backed, we can't keep it in mem or if we do ideally it'd be a mmap'd backed map (which is actually kinda frustrating to implement).

PR #70, opened 2018-05-15
Implement persist/Writer interface

prateek:


persist/writer.go
```diff
@@ -0,0 +1,104 @@
...
+) (MutableSegmentFileSetWriter, error) {
+	return newMutableSegmentFileSetWriter(fs.NewWriter, seg)
+}
+
+func newMutableSegmentFileSetWriter(
+	fn newFSSegmentWriterFn,
+	seg segment.MutableSegment,
+) (MutableSegmentFileSetWriter, error) {
+	w := &writer{
+		fsWriter: fn(),
```
robskillington: Any reason to take a newFn if you call it straight away?  Could just take the writer perhaps?

prateek: Have this ctor to inject mock during tests. The default ctor passes in the public ctor for concrete Writer type.

prateek: talked offline: moved to take the mock directly, instead of the fn indirection.

PR #78, opened 2018-05-16
[search] Optimize conjnuction queries which contain negations

jeromefroe:
This PR optimizes conjunction queries which contain negations by ensuring that we have non-negation queries mixed with negation queries, the non-negation queries are run first and then intersected with the negation queries.


search/query/conjunction.go
```diff
@@ -31,87 +31,126 @@ import (
...
+		default:
+			qs = append(qs, query)
 		}
+	}
 
-		qs = append(qs, query)
+	if len(qs) == 0 && len(ns) > 0 {
+		// We need to ensure we have at least one non-negation query from which we can calculate
+		// the set difference with the other negation queries.
+		qs = append(qs, queries[0])
```
prateek: hm are you guaranteed this itself won't be a negation query?

prateek: also, can you drop the matchall stuff now?

jeromefroe: It certainly could be a negation query but that's ok since we'll execute it like we do normally for negation queries, the matchall and then diff. That's why I didn't want to drop the matchall logic, otherwise you wouldn't be able to execute a negation query by itself.

prateek: ah fair.

search/searcher/conjunction.go
```diff
@@ -78,6 +83,27 @@ func (s *conjunctionSearcher) Next() bool {
 			break
 		}
 	}
+
+	for _, sr := range s.negations {
```
prateek: talked offline: need validation in the conj searcher

